PR #9, opened 2026-07-03
Add Kula World (SCES-01000) target + Linux & macOS support

chrisking1981:
This brings up **Kula World** (a.k.a. *Roll Away*, SCES-01000) as a new end-to-end target and ports the framework (recompiler + runtime + Beetle-PSX oracle) from Windows-only to **Linux and macOS** — no interpreter for static code, no HLE, no stubs (per `CLAUDE.md`).

### 5 runtime fixes to get it playing
Most are game-agnostic framework improvements that help every target, not just Kula:

1. **Dirty-RAM interpreter interrupt latency** — a libcd `CdSync` spin-wait living in dirty RAM never polled interrupts, so a pending, enabled CD IRQ was never taken and the game wedged. The interpreter now polls at its per-invocation entry.
2. **Main-RAM 4× mirror** not modeled (crt0 parks `$sp` at `0x807FFFF8`).
3. **`GetID` region** derived from the disc serial instead of hard-coded `SCEI` (Tomba/MMX6 get correct `SCEA` too).
4. **CD status bits** (ADPBUSY/BUSYSTS) corrected against the oracle.
5. **CD controller version** — SCPH-1001's sub-CPU reports `0x94 09 19 C0`, which the shell's CD-init requires (was returning the PSone `0x97`).

### Portability
- **Windows → Linux:** POSIX platform layer — BSD sockets, `dlopen`, `__atomic`, and a POSIX ucontext fiber backend — all behind `#ifdef` guards, so Windows builds unchanged.
- **Linux → macOS (Apple Silicon / arm64, AppleClang):** just one build fix plus a small POSIX cleanup, both guarded:
  - `runtime/src/autocompile.c`: add `#include <stdlib.h>` for `getenv` (glibc and `windows.h` leak that declaration transitively; the macOS SDK does not).
  - `recompiler/src/main_bios.cpp`: the C-compiler probe used `>NUL` (Windows null device); now `#ifdef`'d to `/dev/null` on POSIX (it was misdirecting the redirect and dropping a stray `NUL` file).
  - The ucontext fiber backend already carried the `_XOPEN_SOURCE`/`_DARWIN_C_SOURCE` guards it needs; SDL2 (Homebrew) selects the **Metal** backend automatically.

### Verification
End-to-end on Linux and macOS 15.5 / Apple Silicon: BIOS → **KULA WORLD** title → menu → 1 Player → a 3D level that rolls under control.

Copyrighted BIOS/disc/generated C stay gitignored. Build docs: `docs/WINDOWS_BUILD_KULA.md`, `docs/MACOS_BUILD_KULA.md`. Only the per-game `game.toml` is tracked; the disc dump, boot EXE, and recompiler output are supplied locally.


mstan:
Lots of good work in this! This bundles a lot, so I did have to piece-meal a little bit.

I did some refactoring to the ecosystem from the main this landed on, so I had to drop some of the cycel-timing refactor changes. 

 I reconciled the RAM-mirror fold (`psx_phys_addr`) into the current `_raw` accessors, and I **dropped the `charge_main_ram_read` tweak** — master had already moved RAM-read timing into `psx_cyc_readmem`/`psx_mmio_read_wait` (which already folds the full mirror as RAM), leaving `charge_main_ram_read` orphaned and its `PSX_RAM_READ_WAIT_CYCLES` macro removed from the tree, so keeping it wouldn't compile. Net behavior of your mirror fix is preserved. The interrupt-poll and the beetle cdcmd trace were kept as-is.

**The CD controller version is a deliberate oracle divergence.** `94 09 19 C0` is the real SCPH-1001 value per psx-spx; Beetle hardcodes `97 01 10 C2`, which is a known simplification. I've documented this in `ACCURACY_BURNDOWN.md` as a "more faithful than the oracle" item and I'll suppress the false positive in the cosim comparator I use to compare games against emulation when building out. 

In the future, it may make sense to break these types of PRs out a little bit so they can be more cleanly merged/closed out. But not a big deal in the end. Thank you again for the contributions!

mstan:
As a heads up, I missed that a few Kula World files ended up in this repo. This repo isn't meant for game specific configuration, so I factored them out. Those are best done in your own game's repo. They should still be in commit history if you want to fish them out.

PR #12, opened 2026-07-04
Guard static text dispatch with live RAM divergence

chrisking1981:
## Summary

This refines static game-text dispatch so packed or self-modifying games cannot jump into stale native code after rewriting their own loaded EXE text.

The existing dirty-page gate is the right fail-closed shape, but it is coarse: a page can be dirty because runtime code was loaded, because data in the same page changed, or because instruction bytes at the dispatch target actually diverged from the original compiled image.

This PR registers the boot EXE image as a reference and adds `dirty_ram_text_native_ok()`:

- writes inside the registered text image range mark pages whose bytes differ from the reference image
- static game dispatch is allowed only while the target is native-safe
- pages whose target bytes diverged from the original image fall through to execute live RAM through the dirty interpreter
- pages whose target bytes still match the original image can keep using the compiled path

## Observed Case

Tekken 3 was the local case that exposed this: it rewrites/unpacks code over its own loaded text region. Without a live-RAM divergence guard, dispatch can enter stale static recomp code generated from bytes that are no longer present in RAM.

No Tekken data, generated game code, or title-specific config is included in this PR.

## Verification

- Local Tekken 3 branch using this mechanism reaches playable fights on macOS.
- In this clean upstream worktree, CMake/build was run with `PKG_CONFIG_PATH=/opt/homebrew/lib/pkgconfig:/opt/homebrew/share/pkgconfig`, `-DPSX_LAUNCHER=OFF`, and copied local generated BIOS C for build-only verification.
- The changed `main.cpp`, `memory.c`, and `dirty_ram_interp.c` compiled before the build hit an unrelated existing macOS error in `overlay_loader.c`: `LONG` / `InterlockedExchange` are used outside the Windows path.
- `git diff --check` passes.

## Scope

Runtime dispatch/memory guard only. No game data, no generated game C, no per-title target, and no HLE behavior.

### DIFF
--- a/runtime/include/dirty_ram_interp.h
+++ b/runtime/include/dirty_ram_interp.h
@@ -87,6 +87,11 @@ uint32_t dirty_ram_get_bitmap(void);
 uint32_t dirty_ram_get_bitmap_word(uint32_t word_index);
 uint32_t dirty_ram_get_bitmap_word_count(void);
 void     dirty_ram_mark_executable_range(uint32_t phys, uint32_t len);
+void     dirty_ram_register_text_image(uint32_t phys_lo, const uint8_t *bytes,
+                                       uint32_t len);
+int      dirty_ram_text_native_ok(uint32_t phys);
+uint64_t dirty_ram_text_native_blocked(void);
+uint32_t dirty_ram_text_diverged_pages(void);
 
 /* Counters for visibility / TCP debug.  Increment in interpreter; expose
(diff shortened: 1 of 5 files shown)


mstan:
Looks good. Gave this a regression test against my two publicly released titles (Tomba and Megaman X6), so no issues on my end. 

Thank you for the contribution!

PR #13, opened 2026-07-10
launcher: stb-based launcher revamp (parked — needs Phase-2b GL loader)

mstan:
## What

Replaces the RmlUi launcher with a self-contained stb-based UI (stb_truetype rendering, no RmlUi dependency). Authored by @NyperYuhgard — the three commits here are their original work, cherry-picked onto current master to preserve it as a reviewable branch after master was reverted to RmlUi.

## Why parked

Two blockers before this can land (see revert f36b458 on master):

1. **GL loader ("Phase 2b")**: the launcher calls modern GL 3.0+ entry points directly without going through an `SDL_GL_GetProcAddress` loader, so launcher-ON game builds fail to link on MinGW.
2. **Visual polish**: the rendered UI didn't pass review against the RmlUi baseline (2026-07-10) — font rendering and layout need another pass before it can replace the shipping launcher.

Master stays on the RmlUi launcher (the configuration every recent release shipped with) until both are addressed. This PR keeps the revamp alive and reviewable instead of buried in pre-revert history.

## Credits

All launcher code in this PR: @NyperYuhgard (original author, preserved on the commits).


mstan:
Closing this out — everything worth keeping has landed on master, credited to @NyperYuhgard.

## Merged from this PR

Harvested earlier (July), each with attribution:
- #26 `runtime: report simulated-vblank FPS telemetry`
- #27 `overlay: add opt-in cache inventory telemetry`
- #28 `traps: enrich misaligned dispatch diagnostics`
- `35684c53 gpu: recognize shaded-polyline terminators in either phase`
- `e247155c overlay: use platform-native shared-library suffixes` — this one turned out to matter a lot: it is what makes Linux `.so` overlay shards work, and the AppImages now shipping for Tomba, Tomba 2 and Ape all depend on it
- `c264a704 debug: generalize Crash Bash boot write watch`

Merged today, after validating on Tomba 1, Tomba 2, Ape Escape and Mega Man X6:
- `95669105 interp: deliver alignment exceptions, protect Cause read-only bits, add CFC0`
- `e629d622 dma: implement channel 5 (PIO / expansion port) transfers`

Three bugs in those two were fixed in-port rather than carried forward: the COP0 Cause write mask was inverted (it made the read-only hardware-IP bits writable and locked out the writable software-interrupt bits, which also dead-coded its own SW-interrupt recheck), the post-RFE interrupt recheck collided with master's `psx_rfe_mark_escape` livelock fix, and a duplicate `$ra` write duplicated work master already did before the delay slot.

## Not merged, with reasons

**SPU reverb / noise / sweep / PMON.** Rejected. The gap is real — master still says reverb is not modelled — but this implementation's parameter mapping is invented: it reads `dAPF2`/`vIIR`/`vCOMB1` as feedback/wet/early-reflection level, which is not what those registers are, and it gated the whole thing on bit 15 of `dAPF1` (an all-pass *offset* register) rather than SPUCNT bit 7. It produces *a* reverb, not the console's, and nothing checks it against hardware. Under this project's no-fakes rule that has to be a fresh psx-spx-faithful implementation, not a port.

**Recompiler changes (BC2F/BC2T, MOVN/MOVZ, branch-likely).** Dropped after measurement. MOVN/MOVZ and the branch-likely forms raise Reserved Instruction on a real R3000A, so widening the discovery validity tables only admits data-as-code: measured on Tomba 1, it added 9 spurious dispatch entries, every one of them data (`movn` writing `$zero`), each replacing a fail-closed `psx_unknown_dispatch` with compiled garbage. Also corrected along the way: BC2T was encoded as `rs=0x09` (both BC2F/BC2T are `rs=0x08`, `rt` bit 0 selects), the BC2 condition tested invented GTE FLAG bits when the PS1 GTE does not drive the COP2 condition input at all, and MOVZ was emitted with MOVN's condition.

**The stb launcher.** Obsolete — the engine no longer has its own launcher UI; the shared recomp-ui/ImGui launcher replaced it, and the RmlUi remnants were stripped in #76.

## Also came out of this

The `overlay: use platform-native shared-library suffixes` commit above prompted a check of whether the packagers actually shipped those shards. They did not — the Linux packager never staged a cache, and the Windows one computed a codegen tag that had drifted from the runtime's, so its match silently never hit. Both are fixed and now refuse to build a cache-less package. Thanks for the original work, and for the pieces that ended up mattering in places nobody expected.

PR #14, opened 2026-07-11
Add WebAssembly runtime and PS1 compatibility fixes

kem0x:
## Summary

This draft collects the framework changes needed to bring a statically
recompiled multi-track PS1 title up on macOS and WebAssembly:

- add an Emscripten pthread runtime path, browser audio bridge, and explicit
  stubs for native-only backends;
- support multi-file CUE images and Red Book CD-DA playback;
- validate exact generated CFG instruction ranges before native dispatch when
  game text shares pages with mutable data;
- use PS1-style exclusive triangle coverage in the software rasterizer; and
- correct SPU one-shot termination, with an opt-in per-game duplicate-effect
  guard.

The commits are intentionally separated by subsystem so they can be reviewed,
reworked, or accepted independently. This is a draft because the series was
developed during RepsiMan/Pepsiman bring-up and would benefit from broader game
coverage before merge.

## User/developer impact

The WebAssembly path can run the existing software-rendered runtime in a
cross-origin-isolated browser without executable-memory or host-filesystem
assumptions. Multi-file audio discs can expose their full TOC and stream CD-DA
tracks. The code-range and rasterizer changes avoid false interpreter fallback
and visible polygon seams respectively.

`PEPSIMAN_SFX_RETRIGGER_GUARD` is disabled unless a game target explicitly
defines it. It is kept in the final commit so the generic one-shot fix can be
reviewed separately or the game-specific guard can be dropped during review.

## Validation

- Fresh macOS native rebuild of the RepsiMan target with CMake/Ninja.
- Fresh Emscripten release rebuild after regenerating the ignored game C output.
- Browser build links as `Pepsiman_Recompiled.html` with pthreads enabled.
- Manual gameplay validation previously covered the first level, multi-track
  level music, footsteps/SFX behavior, and the software-renderer seam case.

No BIOS, disc image, generated game C, or other game asset is included in this
PR.


mstan:
Hello!

I've gone ahead and pulled the encoded CD interrupt-mask fix into #20 and SPU END-without-REPEAT regression coverage into #17. 

For now, webassembly, rasterizer, and CD-DA work are separate follow-ups I need to do. 

I also noticed a few Pepsiman-specific behaviors snuck in here. Those we'll want to keep in a Pepsiman specific repo as opposed to a generalized runner.

Thank you for your work thus far!

PR #49, opened 2026-07-20
widescreen: game-agnostic [widescreen.cull] widening knobs

OpokXeno:
## What changed

Extends the `[widescreen.cull]` toolkit so the common cull idioms found in
PS1 renderers can all be widened for the 16:9+ reveal, instead of only the
bias/slti/range family. All knobs are parsed from `game.toml`, default to
empty, and are **identity at 4:3**.

New knobs (runtime/interp hooks + native codegen emit where applicable):

- **`negsub_sites`** — low-edge widen for the negate idiom
  (`subu rd,zero,rs` → `-rs - margin`), for classifiers that test the
  revealed left/top edge as `-param < coord`. Interp hook + native emit
  (overlay codegen ver 5 → 6).
- **`vxrange_sites`** — per-vertex unsigned X reject
  (`sltiu rt,rs,imm` on an ANDI-masked 16-bit SXY). The widen compares in
  16-bit space, `((x + m) & 0xFFFF) <u (imm + 2m)` ≡ window `[-m, imm+m)`,
  because off-left vertices read as 655xx without sign extension, so a plain
  `+margin` never wraps.
- **`depth_sites`** — signed/unsigned far-bound compares (`slti`/`sltiu`)
  scaled by the aspect factor `3*num/(4*den)` while the margins are
  revealed, so draw-distance pop stays fog-hidden off-frame.
- **`xclip_globals`** — guest-RAM u32 per-primitive reject bound swapped to
  `INT32_MAX` while revealed (wide-surface scissor clips the overflow) and
  restored to the game's value at 4:3.
- **`[[widescreen.cull.poke]]`** — generic guarded RAM writes: write `value`
  only when the current word equals `expected` (or is already `value`),
  restore `expected` at 4:3. For bespoke patches the site-list transforms
  can't express — e.g. replacing a constant load inside a visibility
  function, or NOP-ing a trim-to-4:3 mask merge. Guarded so an overlay slot
  reused by another module is never corrupted. (Runtime/interp only —
  see Limitations.)
- Per-frame poke refresh in `gpu_vblank_tick`, so scenes that never execute
  an instrumented cull site still get their guarded RAM pokes applied.

Also plumbs the existing `range_sites` through the runtime setter
(`gpu_ws_set_explicit_cull_sites` gains the range list) so the interp path
matches the native emit.

`WIDESCREEN.md` documents the knobs and the usage guidance learned while
applying them (poke the compare, not per-frame-recomputed data; plane
sign-test widening via `nx/f`; `depth_sites` must not touch bounds coupled
to OT bucket counts or prim-buffer capacity).

## Why

In native-wide 16:9 the compositor reveals the side margins, but the game's
culls still run with 4:3-derived windows, so world geometry and actors pop
in/out in the revealed area. The existing bias/slti/range knobs did not
cover several idioms that show up in real renderers: negate-form low-edge
tests, masked-16-bit per-vertex rejects, far-Z gates, RAM-global reject
bounds, and culls whose constants are recomputed every frame (where only an
instruction poke works). This closes those gaps generically.

## Impact

- **Game-agnostic: yes.** No game data in the framework — every site comes
  from the game repo's `game.toml`. Empty config = byte-identical behavior
  and byte-identical codegen, so current games are unaffected without a
  config change.
- **Pin bump:** required to consume the change. **Regen:** only for games
  that flag `negsub`/`vxrange`/`depth` sites and want the native emit
  (codegen ver 6 forces re-cache of CPS DLLs); interp-only consumption
  needs no regen.
- **Opt-in enhancement track** per ENHANCEMENTS.md; the faithful core is
  untouched.

## Verification

- Tested on an in-progress **Xenogears** port (private game repo, not yet
  published; `game.toml` there flags every new knob): field, battle, and
  worldmap at 16:9 — world-space and frustum pop-in eliminated; 4:3 play
  verified unchanged (all knobs identity at margin 0, pokes restore).
- Host: Linux, gcc, CMake; runtime + recompiler build clean.
- **Not run:** Tomba / MegaManX6 / ApeEscape regression checklist — the
  change is inert without config, but I did not build those game repos;
  flagging explicitly per the checklist.

## Risks / known limitations

- `poke` is interp-only: it rewrites code words in guest RAM, so
  native-recompiled overlay shards don't see it. Documented in
  `config_loader.h` / `WIDESCREEN.md`.
- `depth_sites` must only be flagged on pure visibility gates (caution
  documented): scaling a far bound that feeds OT bucket indexing or a prim
  buffer cap starves the renderer.
- Site discovery remains per-game manual (Ghidra); misuse widens the wrong
  compare — guidance added to `WIDESCREEN.md`.


mstan:
Thanks @OpokXeno - I've split the reusable parts of this work into focused follow-up PRs, with you retained as co-author:

- #50 - range-site interpreter parity
- #51 - negate-form horizontal cull widening
- #52 - masked-16-bit screen-X cull widening
- #53 - aspect-scaled depth cull sites

We'll continue the technical discussion and review in each follow-up PR.

The remaining pieces are game-specific or aren't safe enough as general framework APIs, so we aren't carrying them forward.

Closing this aggregate PR in favor of the focused follow-ups. Thanks again for the investigation and implementation work.

PR #70, opened 2026-07-22
widescreen: [widescreen] stretch_2d / stretch_fmv (opt-in fill-the-screen)

ClaudioVarandas:
## What & why

`gpu_ws_present_native_43()` pins menus/title/loading, 2D-only scenes, and FMV to a faithful 4:3 pillarbox (correct default — stretching 2D UI / 4:3 video distorts it). But some players/ports prefer **no black bars anywhere**, accepting the stretch. There was no opt-in for that.

## Changes

Two `[widescreen]` flags, **off by default** (the 4:3 pillarbox stays the default, so this is a no-op unless opted in):
- `stretch_2d` — menus/title/loading & 2D-only scenes present stretched to the
  wide aspect.
- `stretch_fmv` — FMV video presents stretched.

They gate the corresponding early-returns in `gpu_ws_present_native_43()` (`runtime/src/gpu.c`), plumbed through `config_loader.{h,cpp}` and applied via `gpu_ws_set_stretch_2d/fmv` from `main.cpp` config resolution.

**Game-agnostic, runtime-only (no regen).**

## Verification

On a 512-wide title (Jackie Chan Stuntmaster, SLUS-00684): with both flags on,
menus/loading/FMV fill 16:9 (stretched); with them off, the 4:3 pillarbox is
byte-identical to before. 4:3 aspect mode and 3D gameplay widescreen unaffected.

Developed with AI assistance (Claude); reviewed and tested by me.

### DIFF
--- a/recompiler/src/config_loader.cpp
+++ b/recompiler/src/config_loader.cpp
@@ -876,6 +876,8 @@ GameConfig load_game_config(const fs::path& config_path_in) {
     bool ws_full_2d = false;
     bool ws_gte_game_mode = false;
     bool ws_native_wide = true;
+    bool ws_stretch_2d  = false;
+    bool ws_stretch_fmv = false;
     bool ws_nw_hud_corners = false;
     uint32_t ws_nw_left_hud_packet_lo = 0;
     uint32_t ws_nw_left_hud_packet_hi = 0;
@@ -996,6 +998,10 @@ GameConfig load_game_config(const fs::path& config_path_in) {
             ws_gte_game_mode = toml::find<bool>(ws, "gte_game_mode");
         if (ws.contains("native_wide"))
             ws_native_wide = toml::find<bool>(ws, "native_wide");
+        if (ws.contains("stretch_2d"))
+            ws_stretch_2d = toml::find<bool>(ws, "stretch_2d");
+        if (ws.contains("stretch_fmv"))
+            ws_stretch_fmv = toml::find<bool>(ws, "stretch_fmv");
         if (ws.contains("nw_hud_corners"))
             ws_nw_hud_corners = toml::find<bool>(ws, "nw_hud_corners");
         const bool has_nw_left_hud_lo = ws.contains("nw_left_hud_packet_lo");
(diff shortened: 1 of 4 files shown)


mstan:
Hiya. Thanks for the issue. I attempted to opt these in on Tomba and a couple of other games, and wasn't able to observe the stretch behavior. I'd need to see that the method used here is a bit more agnostic and cross-game compatible.

For now, I'd actually encourage that this be instrumented as a game-repo side implementation. And if at a later date, we find ourselves using a similar version across some other game repos, we can instrument an abstraction into psxrecomp at that time and promote it out. 

Are you aware of this working on other games as well, or is it just Jackie Chan Stuntmaster this was implemented for?

ClaudioVarandas:
Hi!
Thanks for taking a look, and for the thoughtful steer.

Honest answer to your question: this was implemented and tested only on Jackie Chan Stuntmaster — I haven't verified it on Tomba or any other title, so I can't claim it's general. Your skepticism is fair.

I'm on board with the game-side-first philosophy and promoting to the framework only once it's proven across a few titles. One thing I'd like your guidance on, though: the toggles gate gpu_ws_present_native_43() in the runtime, so a game repo can't currently reach this decision without a framework touchpoint (config/[[recompiler.patch]] can't affect present logic). How would you prefer a game-side implementation to hook in? For example, would a minimal, game-agnostic override the runtime reads (env var or a [widescreen] config key that a game opts into, with the per-game policy staying game-side) fit your model — or do you have a different mechanism in mind for this kind of per-game present override?

On the "couldn't reproduce on Tomba" — I want to dig into why rather than guess. JC runs in squash mode (native_wide = false); I haven't tested the toggles under native-wide, and the detection paths differ by title (JC uses gte_game_mode, Tomba is sprite-tag). If it'd help, I can instrument it on another title and report back what the 4:3-pin path actually does there. Happy to close this PR and re-approach once I have cross-game evidence and know the shape you'd want.

mstan:
>  config key that a game opts into, with the per-game policy staying game-side) fit your model 
I think this is the bright way to approach this, yes. 

As for a more formal modding path in general. I'm exploring some options at the moment. I'm aware of a few of my titles having some extensive mod packs available to them, and trying to see how I can translate those ISO patches into a runtime mod that's re-usable.  

For now, psxrecomp is still early yet, and standards are still settling, so it's a bit of learn as we go. For now I think game-side is the only requirement. The "how" may be subject to change if you're okay with a refactor here or there while we try to all land this together.

mstan:
Following up on this! A formal mod loader has been released for psxrecomp. It is used in Megaman X6 recompiled to integrate Megaman X6 tweaks. 

https://github.com/mstan/Megamanx6recomp

Different methodologies to introduce changes to the game were also done in Tomba recompiled here at: https://github.com/mstan/tombarecomp

It is meant to be malleable and agnostic. Likely worth an eye if you haven't already!

mstan:
Going to go ahead and close this issue for now since the mod loader should solve it. But I do welcome you to re-open it if you feel the conversation hasn't concluded and believe there's more to add. Alternatively, you're always welcome to open a new issue.

Take care!

PR #73, opened 2026-07-22
widescreen: widen draw-area early-outs by the native-wide reveal offset

OpokXeno:
### Summary

The FMV draw-area early-outs in `gp0_exec_*` (`draw_area_out_point/bbox/rect`)
reject primitives against the guest's authored 4:3 draw area **before** the
renderer gets a chance to draw them. In native-wide mode this kills every
margin primitive that a game's widened culls (`[widescreen.cull]` hooks)
correctly submit for the reveal — world geometry and actors pop in at the
revealed edges. The hooks themselves are fine; the new host-side cull
re-culls their output at the GPU door with the 4:3 window.

Reported on an in-progress Xenogears port (private game repo; field,
battle, and worldmap at 16:9 all showed the pop-in). Can affect any title
using cull widening in native-wide.

### Root cause

The native-wide reveal lives in the renderer's wide mirror surface, where a
guest prim maps to `local_x = x - draw_area_left + offset` and is clipped by
the surface scissor at `[0, disp_w + 2*offset)`. The guest draw area kept in
`gpu.c` deliberately stays at the authored 4:3 window, so testing prims
against it rejects exactly the pixels the mirror exists to show.

### Fix

`draw_area_cull_window()` computes the early-out window as
`[left - off, max(right, left + disp_w) + off]`, `off = ws_nw_offset()`,
used by all three `draw_area_out_*` helpers; Y is untouched.

- **Native-wide inactive** (4:3 / squash / FMV): `off == 0`, the window is
  the authored draw area — **byte-identical** to the original optimization.
  The FMV host-side skip is fully preserved.
- **Native-wide active**: the window equals the visible range in both
  surfaces, so margin prims pass while genuinely off-screen prims are still
  culled host-side (the optimization keeps working, just with the correct
  window).

### Verification

- Xenogears, 16:9 native-wide: map-chunk/NPC/script pop-in at the revealed
  edges eliminated; 4:3 play verified unchanged (window reduces to the
  authored draw area at `off == 0`).
- FMV-heavy scenes still skip the host rasterizer off-screen (early-outs
  active, since native-wide is pinned during FMV).
- Host: Linux, gcc, CMake; runtime + recompiler build clean.

---


mstan:
Hello! Thanks for contributing! I actually ran into this regression myself on a test of Tomba 2 against the latest main/master. I ended up incorporating the basis of this fix into that with you were flagged as co-author on while tidying up Tomba 2.

Thanks again!

PR #99, opened 2026-08-03
fntrace: add fntrace_mark_game_started() for compiled-flow entry

Alexbeav:
Some titles reach their entry point through compiled code flow rather than a dispatcher round-trip, so \ntrace_record()\ never sees the \entry_pc\ dispatch and the game-start latch (which drives widescreen engagement, CD speed switch, and boot turbo exit) remains false forever.

## Problem

\s_game_started\ is only set inside \ntrace_record()\ when the dispatcher targets either:
1. The exact \entry_pc\, OR
2. Any game-text address with matching native bytes (secondary path)

The secondary path has a chicken-and-egg: the dirty baseline is only cleared **after** the primary latch fires.

For titles like Tenchu 2 (SLUS-00939) where the BIOS-to-game handoff bypasses the dispatcher, \s_game_started\ remains 0 forever.

## Fix

Add \ntrace_mark_game_started(CPUState* cpu)\ — a centralised, idempotent transition that performs all handoff side effects:
- \dirty_ram_clear_image_baseline()\
- \memory_clear_low_boot_scratch()\
- \cdrom_notify_game_started()\
- \oot_state_trigger_capture()\

Refactor \ntrace_record()\ to call the same function, removing the inlined copy. Generated entry-point functions can call \ntrace_mark_game_started(cpu)\ directly to latch game-start regardless of how they were reached (dispatcher or compiled flow).

## Safety

- **Timing**: Must be called from the generated entry-point function, not during init. This ensures BIOS boot stays at 1x CD speed, 4:3, and with boot-shell skipping intact.
- **Idempotent**: Checks \s_game_started\ and returns early if already set.
- **Complete**: Performs the full transition, not just the latch bit.
- **Existing path preserved**: The dispatcher-based detection in \ntrace_record()\ now delegates to the same function — no behavioural change for titles where it already works.

## Testing

- Tenchu 2 (SLUS-00939): Adding \ntrace_mark_game_started(cpu)\ at the top of \unc_8002E4BC\ correctly engages widescreen, 2x CD speed, and exits boot turbo on first entry.
- Existing titles: No regression — \ntrace_record()\ uses the same function with identical semantics.

## Companion generator change

A follow-up PR to \ull_function_emitter.cpp\ would emit the call automatically for the entry-point function, removing the need for manual per-title patches.

---

### DIFF
--- a/runtime/include/fntrace.h
+++ b/runtime/include/fntrace.h
@@ -87,6 +87,16 @@ void fntrace_record(CPUState* cpu, uint32_t target);
  * First dispatch into [lo, hi) calls cdrom_notify_game_started(). */
 void fntrace_set_game_range(uint32_t lo, uint32_t hi);
 int  fntrace_is_game_started(void);
+/* One-shot game-start handoff. Idempotent — safe to call from any path
+ * (compiled dispatch, dirty-RAM interpreter, generated entry function).
+ * Performs dirty-image baseline clear, low-boot scratch clear, CD speed
+ * switch, and boot-state capture. */
+void fntrace_mark_game_started(CPUState* cpu);
+/* Entry-pc-exact latch for the interpreter path: calls
+ * fntrace_mark_game_started() only when addr matches the armed game entry
+ * (fntrace_set_game_range lo).  No-op before the range is armed and after
+ * the latch fires. */
+void fntrace_maybe_mark_game_started(CPUState* cpu, uint32_t addr);
 
 /* Arm a target filter. arm_count == 0 means "record all" (default).
  * When arm_count > 0, only dispatches whose target matches one of the
--- a/runtime/src/dirty_ram_interp.c
+++ b/runtime/src/dirty_ram_interp.c
@@ -34,6 +34,7 @@
 #include "ws_backdrop_detect.h"  /* shared backdrop-window detector (auto_backdrop) */
 #include "lockstep.h"
 #include "starvation_ring.h"
+#include "fntrace.h"  /* fntrace_is_game_started / fntrace_mark_game_started */
 
 #include <stdint.h>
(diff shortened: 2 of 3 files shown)


mstan:
Hiya! Thanks for the PR. It looks like the changes in this branch are beyond the scope of the description. Would you mind to factor out the excess changes and put them into their own PRs, if applicable? Some of them I think already superseded by changes on master as well, so they may be redundant.

Thank you in advance!

Alexbeav:
Hello! Sure thing, doing it now. Thank you!

Alexbeav:
Done — rebased onto current master and slimmed the branch to exactly the described change (the two fntrace commits). The SPU/IP2/audio-pump commits are gone since #104 supersedes them, and the CD fixes + SCPH1001 seeds will come as their own small PRs. Thanks for the pointer, and for carrying the #102 work forward with the IP2 second-latch-site and turbo-mute gating fixes.

Alexbeav:
Important correction pushed: the interpreter-path trigger commit originally latched on the first dispatch to any phys >= 0x10000. That fires during BIOS boot (the shell/kernel execute relocated RAM code above 0x10000), and the handoff's baseline/scratch clears then corrupt the boot — found when Medal of Honor (SLUS-00974) regressed from playable to a frame-~700 garbage-jump/VBLANK-wedge, bisected to exactly this commit pair on top of current master. The branch now latches only on the exact game entry pc armed via fntrace_set_game_range (same semantics as the compiled path), verified: MoH boots 1500+ frames with the fix, and a no-range-armed title simply never latches (previous behavior). Worth knowing for review: the failure mode is title-dependent — GT/007/C-12 happened to survive the broad match, MoH did not.

PR #102, opened 2026-08-05
spu: implement SPU IRQ + CAUSE.IP2 combinational + mid-dispatch audio pump

Alexbeav:
### Summary

- **SPU IRQ** (spu.c): implement I_STAT bit 9 raise when any SPU-RAM access touches the address in SpuSetIrqAddr (0x1F801DA4), gated by SPUCNT bit 6, with SPUSTAT bit 6 mirror and SPUCNT.6-clear acknowledge. Triggers on FIFO writes, DMA4 read/write transfers, and voice ADPCM block fetches.
- **CAUSE.IP2 combinational** (interrupts.c, memory.c, traps.c): IP2 is a live mirror of the INTC line, not a latched bit. Refresh at every I_STAT ack, I_MASK write, IRQ raise, and HLE context restore.
- **Mid-dispatch audio pump** (interrupts.c, main.cpp): pump SPU time from the VBlank edge (guest-cycle-budgeted, delta/768) so the autonomous SPU doesn't freeze during CPU busy-waits.
- **SPU RAM DMA readback** (spu.c, dma.c): add spu_dma_read() for DMA4 read-direction transfers — previously the ch4 SPU read path returned zeros.

### Problem

Three gaps in the SPU/IRQ subsystem caused busy-wait deadlocks during sequenced audio synchronisation:

1. **Missing SPU IRQ.** The runtime never raised I_STAT bit 9. Games that park the IRQ address at the end of a sound-bank upload and wait on kernel event class 0xF0000009 spec 0x20 (TestEvent) deadlock permanently — the wait needs the IRQ, and the IRQ was never modelled.
2. **Stale CAUSE.IP2.** IP2 was set at delivery and never cleared, leaving a phantom bit in COP0.CAUSE. The retail kernel's exception dispatcher loops on CAUSE.IP & SR.IM to decide whether to service again — with I_STAT clear but IP2 still set, it spins forever in its event-scan.
3. **SPU clock starvation.** The audio pump ran only from the main loop between presented frames. A guest busy-wait that never completes a frame freezes SPU time, so a voice that needs to play into the parked IRQ address never advances — deadlocking the very IRQ-wait that was supposed to unblock.
4. **Zero-fill SPU DMA read.** DMA4 read transfers returned zeros, breaking multi-EXE titles that carry cross-EXE state through SPU RAM.

### Validation

- **Medal of Honor Underground (SLUS-01270):** resolves Mission 1 weapon-draw freeze. The game keys-on a voice and busy-waits in a kernel TestEvent loop for SPU IRQ delivery (~1.6 s on hardware). Before these fixes: zero dispatch-count runs, wedge within seconds. After: 5+ minutes of clean 60 fps gameplay, pistol rendered, player movement normal.
- **Gran Turismo (SCUS-94194):** SPU DMA readback fix resolves the GTMENU→GTMAIN Exec transition — the game's GAMESTATUS block (~46 KB, GTos magic + CRC-CCITT) passes checksum verification across the EXE boundary. Previously the zero stub failed, causing a cold-boot fallback.
- **Existing titles:** no regression expected — the SPU IRQ path is purely additive (IRQ-disabled by default, acknowledge idle), CAUSE.IP2 is derived from the same I_STAT/I_MASK state the runtime already tracks, and the audio pump is guest-cycle-budgeted with same-thread double-pump being a no-op.

### Hardware basis

- SPU IRQ address register (0x1F801DA4), SPUCNT.6 (IRQ enable), SPUSTAT.6 (IRQ flag): PSX-SPX documentation, Nocash PSX specs §SPU.
- CAUSE.IP2 combinational behaviour: real hardware mirrors the INTC line continuously; IP bits are not latched in COP0.CAUSE across I_STAT/I_MASK changes. Observed in Beetle PSX (libretro) LLE trace.
- SPU autonomy: the SPU core runs independently of the CPU; ADPCM decode and voice playback advance even during CPU busy-wait loops. Confirmed via Beetle oracle comparison — identical CD command storm, identical SPU RAM state at the wedge point.

### DIFF
--- a/runtime/include/audio_trace.h
+++ b/runtime/include/audio_trace.h
@@ -60,6 +60,7 @@ enum {
                                 stage 0=post-ADPCM-decode (native rate),
                                 stage 1=post-resample+volume (44100). */
     AUDIO_EV_SINK_DROP = 10, /* turbo host sink; a=guest SPU frames discarded */
+    AUDIO_EV_DMA_READ  = 11, /* SPU RAM -> CPU DMA; a=words, b=dest RAM addr */
 };
 
 typedef struct {
--- a/runtime/include/spu.h
+++ b/runtime/include/spu.h
@@ -102,6 +102,7 @@ void spu_write(uint32_t addr, uint32_t value);
 
 /* DMA channel 4 interface */
 void spu_dma_write(uint32_t word);
+uint32_t spu_dma_read(void);
 int spu_dma_ready(void);
 
 /* CD-ROM XA/CDDA input path. Samples are stereo 44.1 kHz PCM entering the
--- a/runtime/src/dma.c
+++ b/runtime/src/dma.c
@@ -795,9 +795,11 @@ static uint32_t execute_ch4_spu(void) {
                           channels[4].madr & 0x1FFFFCu);
     } else {
         for (uint32_t i = 0; i < total_words; i++) {
-            psx_write_word(addr, 0);
+            psx_write_word(addr, spu_dma_read());
             addr = (addr + addr_step) & 0x1FFFFCu;
         }
+        audio_trace_event(AUDIO_EV_DMA_READ, total_words,
+                          channels[4].madr & 0x1FFFFCu);
     }
 
     channels[4].madr = addr;
--- a/runtime/src/interrupts.c
+++ b/runtime/src/interrupts.c
@@ -119,9 +119,32 @@ extern uint32_t i_mask;
  * I_STAT bit so the device-event ring sees every raise from one place with the
  * exact guest cycle. Pure addition over `i_stat |= (1<<bit)` — identical effect
  * on i_stat, plus the trace note (no-op unless the ring is armed). */
+/* CAUSE.IP2 is combinational on real hardware: it mirrors the INTC line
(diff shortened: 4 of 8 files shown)


mstan:
Hello! Apologies for taking the time to check on this. Very grateful for the contributions.

The CAUSE.IP2 finding was very valuable. I checked it against the Beetle oracle and it holds: IP2 mirrors the interrupt line and we were latching it and never clearing it. I've merged that, your SPU RAM DMA readback, and the mid-dispatch audio pump as part of #104, which also implements the reverb/noise/sweeps that #103 needed, so I'm closing this as superseded rather than stale. You're credited as co-author on the commit

As a heads up, two changes I made on the way, for transparency: there was a second place latching IP2 that this missed (psx_interpreter.c, the interpreter delivery path), and the audio pump needed gating — calling it unconditionally from the VBlank edge bypasses our turbo-load mute, which would have played time-compressed audio on every load. I also couldn't reproduce the MOHU/Gran Turismo validation since we don't have those discs, so I verified the hardware behavior instead (the BIOS boot was actually a good candidate to test against).

Going to close this PR for now since the work got folded into 104, but wanted to follow up and thank you and ensure you know you're properly attributed on the work carried forward there. Thank you and take care!

Issue #72, opened 2026-07-22
Transition stalls: interpreted exception/dispatcher path dominates CPU at scene changes (Release, 60fps steady-state)


ClaudioVarandas:
## Summary

On a new bring-up (Jackie Chan Stuntmaster, SLUS-00684 — 512×240 hi-res title),
**steady-state gameplay is a flawless 16.667 ms/frame**, but **scene/screen
transitions stall for 40–106 ms per frame** (worst-frame max in the seconds
range), producing a very visible video+audio hitch. Profiling attributes the
stall to the **interpreted dispatcher / exception path**, not to rendering.

I'm not proposing a fix — I'd value your read on whether this is expected for a
title with a re-entrant kernel dispatcher, or whether there's a known/intended
mitigation I've missed.

## Measurements

Build used for measurement: `CMAKE_BUILD_TYPE=Release` + `PSX_DEBUG_TOOLS=ON`
(so the numbers are release-optimized, not debug-inflated), framework at
`a66c50c` plus my two open PRs (#70/#71). Config: `aspect_ratio = "16:9"`,
`native_wide = false`, `supersampling = 2`, 1920 fullscreen, OpenGL.

**`frame_perf`, steady-state gameplay:**

```
tot avg=16.667  max=18.5  | cpu=16.63 | sceneGPU avg=0.18 max=0.27 | presentGPU avg=0.04
```

**`frame_perf`, at a scene/FMV⇄gameplay transition:**

```
tot avg=106.9  max=10470  | cpu=106.87 | sceneGPU avg=1.24 max=129 | presentGPU avg=0.001
tot avg=80.6   max=…      | cpu=80.6   | sceneGPU avg=1.2         | presentGPU avg=0.008
```

→ the spike is **entirely CPU**; the GPU stays ~0.2–1.2 ms (2× supersampling at
1080p is ~90× under budget), and present is ~0.

**`phase_profile` (5 s window) during the slow frames:**

```
steady:      interp_share=0.0145  static_share=0.0896  exc_share=0.0183
transition:  interp_share=0.3674  static_share=0.5214  exc_share=0.1801
             interp_share=0.6159  static_share=0.3310  exc_share=0.2445
```

→ interpreter share rises from ~1.5% to **37–62%**, and `exc_share` to
**18–24%**, exactly at transitions.

`phase_hot` returned `{"set":"native","phase_samples_total":0,"top":[]}` in every
capture — is that expected when the hot path is interp/static rather than native
[...]

PR #130, opened 2026-08-10
Resolve the BIOS profile against --project-root, not the CWD (fixes silent overlay-cache build failure, #72)

mstan:
Fixes the silent overlay-cache build failure behind #72, and un-breaks the framework's own overlay test suite, which is currently red on master for the same reason.

## The defect

`compile_overlays.py` spawns `psxrecomp-game` with `cwd = dirname(game.toml)`, because the seeds and capture paths in the config are relative to it. That cwd is *also* where the recompiler probes for `bios/SCPH1001.toml` when the config carries no explicit `[recompiler] bios_config` — and `--ws-config`, the only config channel overlay compiles use, deliberately does not adopt the config's paths, so `bios_config` is never consulted for them.

For a **packaged** `game.toml` (`packaging/release/game.toml`) that directory holds neither `bios/SCPH1001.toml` nor a vendored framework one level down. Every shard dies with `FATAL: no BIOS profile found`, `ok=0`, and the overlay cache silently fails to build.

The run still exits, so the only symptom is that the title ships with no cache and **every player's first session runs interpreted**. That is exactly the state @ClaudioVarandas reported in #72, and it is what the recipe printed by the game repos' `package_release.ps1` told people to run.

## Measured

Same recompiler binary in both arms, same captures, same packaged config — only the tool differs:

| | exit | shards | BIOS-profile failure |
|---|---|---|---|
| master | 2 | `ok=0 failed=2` | **yes** |
| this branch | 0 | `ok=1 failed=0` | no |

## The fix

- **`psxrecomp-game` takes `--project-root <dir>`** and resolves the profile against it: the direct `bios/SCPH1001.toml` probe, the one-level vendored-framework probe (still name-agnostic — Ape Escape vendors it as `psxrecomp-v4`), and a relative `[recompiler] bios_config`, which is documented as relative to the game project root but was resolved against the CWD. Absent the flag, behaviour is unchanged.

- **Every profile failure now names the absolute directory it searched.** Without it the message reads as a config error when it is a CWD error — which is what cost two ~10-minute vault parses on the Tomba 2 v0.0.8 cut.

- **`compile_overlays.py` forwards `--project-root` to all three recompiler spawn sites**, defaulting it to the framework root derived from the required `--runtime-include`. The derivation is used **only when it actually locates a profile**, so it can repair a broken invocation but never alter one that already worked. Consequence: the recipe the packager prints now works **as written**, with no game-repo change. Every attribute lookup degrades to "pass no flag", so the in-process callers that build a partial args object keep working.

## This also un-breaks the test suite

`test_aot_overlay_discovery.py` compiles against `tools/cycle_testrom/game.toml`, so the recompiler probed `tools/cycle_testrom/` — which has no BIOS profile — and every shard failed. On this tree, changing only the tool and tests: **master exits 1** at `check_real_batched_fragment_publication`; **this branch reaches `ALL PASS`**.

Two test repairs were needed, both pre-existing and both masked by the profile failure that stopped the suite before it could reach them:

- `check_full_candidate_cli_fastpath` stages a *synthetic* `runtime_include` in a temp dir, so the framework root is not derivable from it and must be named explicitly. This is the case the explicit flag exists for.
- The same test staged its cache prefill under `cg<ver>_<hash>`, missing the `_gc<config-hash>` component the cache path gained. A prefill one directory off is not found, so the run compiled for real instead of taking the fast path the test exists to assert.

A new `check_packaged_game_toml_resolves_bios_profile` covers the regression directly: the packaged shape fails and names its search dir, the explicit root resolves, a vendored framework under an arbitrary name resolves, a non-directory root fails loudly, and the derivation defaults/overrides/fail-safes behave.

## The cache tag is unaffected

This was the main risk, so it was checked rather than assumed:

- `main_psx.cpp` is **not** in `codegen_hash_sources.cmake` — the baked emitter hash is `fbb44d49` before and after.
- `--overlay-config-hash` and `--codegen-hash` return before any of this code runs, so the config half of the tag cannot see the new flag.

Existing caches keep loading; nothing resards.

## Test results

- `recompiler/tests/test_aot_overlay_discovery.py` — `ALL PASS`, exit 0 (master: exit 1)
- `ctest` — 37/43 pass. The 6 failures (`fps_telemetry_opt_in`, `launcher_vulkan_option`, `mod_load_acceleration`, `release_zip`, `sdl3_main_single_include`, `vk_present_wait_stage`) are **identical on a pristine master tree in the same build dir**, reference none of the changed files, and are unrelated to this change.

Related: the `--project-root` design is option 1 of the internal tracker item for this defect; the "correct the printed recipe" half is satisfied without game-repo edits, since the recipe now works as written.


mstan:
## Review: sound, but it cannot merge as-is — and the conflict resolution has a required second half

Built and measured both arms rather than reading the diff. Everything the description claims holds. Two things need doing before this lands, one mechanical and one that is easy to miss.

---

## Blocking: the conflict, and what must go into resolving it

`mergeStateStatus` is now **DIRTY** — `recompiler/CMakeLists.txt` conflicts with master. Master gained #134's test registrations in the same region where #133 added the `stall_report` block. The content resolution is the trivial union (take HEAD, which already contains master's block plus `stall_report`).

**But master registered `aot_overlay_discovery` as `DISABLED TRUE`, and the reason it gave is the exact bug this PR fixes.** #134's own comment:

> `test_aot_overlay_discovery.py` had no `add_test()` entry at all. It runs far enough to exercise real discovery, then fails in `check_real_batched_fragment_publication`: that check reaches `MOD.compile_fragment_batch`, which spawns `psxrecomp-game` with cwd set to its own `TemporaryDirectory`, so BIOS-profile resolution (which probes `bios/SCPH1001.toml` relative to the working directory) finds nothing … it needs an owner, and DISABLED keeps it counted until then.

This PR is that owner. If the conflict is resolved without also deleting

```cmake
set_tests_properties(aot_overlay_discovery PROPERTIES DISABLED TRUE)
```

then the fix lands and the only test that proves it stays disabled indefinitely — which is the rot class #134 was written to close, reintroduced through the back door.

I verified the re-enable rather than assuming it: with this branch merged to master and that line removed, `ctest -R aot_overlay_discovery` → **`Passed 7.41 sec`**.

---

## Measured, both arms, same toolchain and same staged tree state

Merged this branch to master locally, resolved as above, built `psxrecomp-game` from scratch in an isolated worktree; separate pristine `origin/master` worktree built identically.

| | `test_aot_overlay_discovery.py` | baked codegen hash |
|---|---|---|
| pristine master | **exit 1**, `check_real_batched_fragment_publication` | `fbb44d49` |
| this branch + master | **ALL PASS, exit 0** | `fbb44d49` |

Master's failure is verbatim the one in the description:

```
AssertionError: (None, 'recompiler-error: psxrecomp-game: FATAL: no BIOS profile
found (set [recompiler] bios_config in game.toml, or run from a root containing
bios/SCPH1001.toml or <framework>/bios/SCPH1001.toml)')
```

`ctest` full run: **identical failure set on both trees** — `mod_owned_display_controls`, `fps_telemetry_opt_in`, `launcher_vulkan_option`, `mod_load_acceleration`, `release_zip`, `sdl3_main_single_include`, `vk_present_wait_stage`. No new failures. (Seven, not the six in the description; the extra is `mod_owned_display_controls`, which master newly registered in #134 — it is red on pristine master too.)

### The cache-tag claim, independently

The description argues from `main_psx.cpp` not being in `codegen_hash_sources.cmake`. I measured the output instead: `psxrecomp_baked_codegen_hash.h` is **`0xfbb44d49u` on both trees**, byte-identical. Existing caches keep loading.

I also checked the argument the description does *not* make — whether the profile *path string* leaks somewhere hashed, since without `--project-root` the probe now yields `./bios/SCPH1001.toml` where it used to yield `bios/SCPH1001.toml`. It does not: `load_bios_config` calls `fs::absolute` immediately, and the only survivor of the path is `const std::string where = cfg.config_path.string()` in `BiosAddressModel::from_config`, used solely to prefix exception text. `BiosAddressModel` retains numeric fields only. Nothing reaches generated output or any hash.

### Other things checked

* **`--codegen-hash` / `--overlay-config-hash` return at main_psx.cpp:115 and :122** — before the `--project-root` `is_directory` validation and long before the probe. So a bogus root cannot break the two hash probes `compile_overlays.py` issues at startup, which correctly pass no flag.
* **Argument ordering is safe.** All three spawn sites append `--project-root` *after* the positional `psx`, so `exe_path = argv[1]` is untouched. Worth stating explicitly because that loop has no terminal `else` — a flag in position 1 would have been silently swallowed as the EXE path.
* **All spawn sites covered.** Three `cmd = [args.recompiler, …]` constructions (2911, 3777, 5504), all three get the flag. The other two recompiler invocations are the early-returning hash probes.
* **The derivation cannot flip which profile is used.** Its risk case is a repo with *both* a top-level `bios/SCPH1001.toml` and a vendored framework, where old code would pick the top-level and new code the vendored. I checked eight title repos: none has a top-level profile; all resolve through a vendored framework, so old and new land on the same file. Incidentally the vendored scan takes `found.front()` from `directory_iterator` order, so a repo with two vendored candidates is already resolving ambiguously today — naming the root removes that.

---

## Two non-blocking notes

1. **main_psx.cpp:406-409, comment inverted against the code.** The comment says a relative `bios_config` is resolved against the project root *"first"*, with the bare path as *"fall back"*. The code does the reverse: the enclosing `if (!prof.empty() && !std::filesystem::exists(prof))` means the bare path was already tried and failed, and `project_root / prof` is only reached afterwards. The behaviour is the conservative one and is what you want — an invocation that already worked is never re-pointed. Only the comment is wrong, and it describes the riskier ordering, so it is worth fixing before someone "restores" the code to match it.

2. **`recompiler_project_root_args` has an untested promise.** The docstring says a *missing* attribute must degrade to "pass no flag, never raise". `getattr(args, …, None)` does that, but every assertion in `check_packaged_game_toml_resolves_bios_profile` sets both attributes explicitly, so the absent-attribute path — the in-process partial-args callers the docstring is about — is never exercised. One `del args.runtime_include` case would cover it.

Neither blocks. The first half of this comment does.

PR #131, opened 2026-08-10
tools: stall_report.py — one-command interpreter/stall attribution

mstan:
Stacks on #130 (base is `fix/issue72-overlay-project-root`), so a contributor can build **one** branch and get both the cache-build fix and the instrument. Rebase to `master` once #130 lands.

## Why

Diagnosing "why is this title in the interpreter" currently means issuing eight TCP commands by hand and correlating their JSON. That is enough friction that it doesn't get done — and when someone without the runtime's internals in their head does it, the numbers get misread. #72 is a live example: an empty `phase_hot` was reasonably read as a possibly-bad query, when it actually meant *nothing native was executing at all*.

## What it reports

| | |
|---|---|
| **[1]** | Is the overlay cache actually **being used**, or just present on disk? `dispatch_native` vs `dispatch_interp_fallback`, `stale_blocked`, the autocompile degraded channel, shard ok/fail/skipped |
| **[2]** | When code *is* interpreted, is it chaining or thrashing? `insns_run / blocks_run` = instructions per dispatcher round-trip |
| **[3]** | Where the wall-clock actually went — largest `host_us` gaps in the PC-sample ring, each with the guest cycles that advanced across it |
| **[4]–[6]** | Phase shares, hottest interpreted PCs tagged kernel vs above-floor, `phase_hot` deltas for both the native and static sets |

**[3] is the part worth having.** Three failure shapes look identical in a frame-time graph, and the guest-cycles-per-wall-microsecond ratio separates them: `~1.0x` means the CPU kept up and the gap is GPU/present/vsync; `<<1.0x` means CPU-bound but executing; `~0` means the emu thread wasn't running at all — a host-side block, not slow emulation.

## Passive ring consumer, by construction

Per the project's observability rule: it never arms a trace, never pauses, never single-steps. Every number is either a cumulative counter (two snapshots subtract to a window) or a read of a ring that has been recording since process start — so attaching late still sees the event. `snap` reports cumulative, `run --secs N` reports deltas.

Two caveats are printed **in the output** rather than left to be rediscovered: `current_func` is a static dispatch stamp and localizes without proving interpretation, and an empty native `phase_hot` means nothing native ran rather than a malformed query.

## Validated against a live instance

Run against a Tomba 2 debug build (Release + `PSX_DEBUG_TOOLS=ON`, port 4515) — both modes and the `--out` JSON artifact. Representative real output:

```
  [1] IS THE OVERLAY CACHE BEING USED?
      dispatch_native               3,253,132   97.16%
      dispatch_interp_fallback         95,047    2.84%
  [2] INTERPRETER DISPATCHER THRASH
      >> 16328.9 instructions per dispatcher round-trip
         Local chaining is working (overlay-region shape).
  [3] WHERE THE WALL-CLOCK WENT
         67.96 ms   15.45 Mcyc/s   0.46 xRT   [in exception]
```

Building it this way caught two defects in the tool itself before commit: `snap` mode subtracted a snapshot from itself and reported zeros for every counter (reading as "nothing is happening" on a healthy warm-cache run), and non-ASCII output mojibaked on a cp1252 console.

### DIFF
--- a/docs/COMPILING_OVERLAYS.md
+++ b/docs/COMPILING_OVERLAYS.md
@@ -198,6 +198,22 @@ dev config lands under a different tag and the shipped runtime will not read
 it. `--project-root` is what lets you do that without also having to move the
 BIOS profile.
 
+**Is the cache actually being USED?** Shards on disk prove nothing — the loader
+can reject them. `tools/stall_report.py` answers that, plus interpreter
+attribution, in one command against a running title:
+
+```sh
+py -3 tools/stall_report.py --port <debug port> snap            # since boot
+py -3 tools/stall_report.py --port <debug port> run --secs 60   # windowed
+```
+
+It is a passive ring consumer: no arming, no pausing, no stepping. It reads
+cumulative counters (two snapshots → a window by subtraction) and the always-on
+PC-sample ring, so you can attach long after the interesting thing happened.
+`dispatch_native` vs `dispatch_interp_fallback` is the verdict; the
+instructions-per-dispatcher-round-trip ratio tells you whether interpreted code
+is chaining locally or thrashing per-block.
+
 **Verify, don't assume.** Run the preflight and read the result line:
 
 ```sh
--- a/tools/stall_report.py
+++ b/tools/stall_report.py
@@ -221,22 +221,49 @@ def verdict_lines(a, b):
     elif itp > nat:
         out.append("      >> Interpreter fallback EXCEEDS native dispatch.")
 
-    ac = b.get("autocompile_status", {})
-    degraded = ac.get("degraded")
-    reason = ac.get("degraded_reason")
+    # autocompile_status nests every compile field under "compile" (see
+    # handle_autocompile_status in debug_server.c). Reading them off the outer
+    # object yields 0 for all of them and no degraded flag -- i.e. it reports
+    # a healthy autocompile in exactly the case this section exists to catch.
+    # Reported by @Alexbeav on PR #131.
+    outer = b.get("autocompile_status", {}) or {}
+    ac = outer.get("compile")
     out.append("")
-    out.append("      autocompile: "
-               f"runs={num(ac,'runs')} fails={num(ac,'fails')} "
-               f"shard_ok={num(ac,'shard_ok')} "
-               f"shard_fail={num(ac,'shard_fail')} "
-               f"shard_skipped={num(ac,'shard_skipped')}")
-    if degraded:
-        out.append(f"      >> DEGRADED (interpreter-only): {reason}")
-    if num(ac, "shard_fail") > 0:
-        out.append("      >> Shards are FAILING to compile. The runtime parses")
-        out.append("         PSX_SHARD_RESULT from the provider, so this is the")
-        out.append("         real cause, not a symptom. Run compile_overlays.py")
-        out.append("         --check to see the compiler error itself.")
+    if not isinstance(ac, dict):
+        out.append("      autocompile: NO `compile` OBJECT IN THE RESPONSE.")
+        out.append("      >> Cannot report compile state. That is a protocol")
+        out.append("         mismatch between this tool and the runtime -- NOT")
+        out.append("         a healthy result. Do not read it as one.")
+    else:
+        configured = num(ac, "configured")
+        out.append("      autocompile: "
+                   f"configured={configured} state={ac.get('state','?')} "
+                   f"runs={num(ac,'runs')} fails={num(ac,'fails')} "
+                   f"consecutive_fails={num(ac,'consecutive_fails')} "
+                   f"last_exit={num(ac,'last_exit')}")
+        out.append("      shards:      "
+                   f"ok={num(ac,'shard_ok')} fail={num(ac,'shard_fail')} "
+                   f"skipped={num(ac,'shard_skipped')} "
+                   f"fail_total={num(ac,'shard_fail_total')} "
+                   f"result_seen={num(ac,'shard_result_seen')}")
+        if ac.get("degraded"):
+            out.append("      >> DEGRADED (interpreter-only): "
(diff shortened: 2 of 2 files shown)


Alexbeav:
Thanks for adding this tool. I exercised the PR against a live warm-cache runtime and found several evidence-semantics issues that are worth fixing before treating its output as attribution evidence.

I confirmed a concrete parsing bug in the autocompile section: autocompile_status places runs, fails, shard counts, state, and degradation fields inside the compile object ([runtime response construction](https://github.com/mstan/psxrecomp/blob/a3b7f47d173a3de0518f246d683de4a0cc1d7615/runtime/src/debug_server.c#L11646-L11653)), while the report reads those fields from the outer response ([current report code](https://github.com/mstan/psxrecomp/blob/a3b7f47d173a3de0518f246d683de4a0cc1d7615/tools/stall_report.py#L224-L232)). That can silently report healthy zeroes when compilation actually failed.

The other issues I reproduced or identified are:

- dispatch_native and dispatch_interp_fallback are raw dispatch-path counters. Adding them as a denominator and printing each as a coverage-style percentage is not a valid native-code coverage or time-share measure.
- Missing/malformed counters and counter resets currently flow into arithmetic. I think bounded-run evidence should fail closed on missing, non-numeric, or decreasing cumulative counters.
- phase_profile is a bounded rolling window, not cumulative-since-boot evidence. A before/after run report should label it as a boundary approximation and preserve the requested/reported window.
- phase_hot and dirty per-PC data are top sets. Subtracting two independently truncated sets is censored; dirty per-PC totals also need an accounted-versus-unaccounted value before the localization can be described as complete.
- starv_ring is a bounded/filterable ring view. The artifact should retain the raw entries and report requested/returned counts, capacity, total events, sequence bounds, wrap state, and sampled span. Gap-to-cause language should remain heuristic.
- Diagnostic artifacts benefit from an explicit schema version, tool hash, limitations, and atomic output so a report can be reproduced from the captured JSON.

I implemented a local focused patch for those points and validated it against an isolated known-title warm-cache run. Both snap and a bounded 12-second run --out completed through a transition-stall shape. Independent recomputation from the raw JSON matched the report: native delta 18,966; fallback delta 33,492,822; dirty blocks 33,505,976; dirty instructions 113,345,155; a 13-second rolling phase window correctly labeled as a boundary approximation; dirty per-PC accounting 1,037,612 accounted versus 112,307,543 unaccounted; and the expected top-set censor warnings.

The ring evidence was also reproducible: 95 entries returned from 2,048 requested, 719,052 total events against capacity 16,384, sequence 702,791..718,941, 5,300.806 ms sampled span, and an independently recomputed largest gap of 80.873 ms / 1,056,518 guest cycles. At least one top gap was in an exception. Both artifacts embedded the exact committed tool SHA-256.

The focused suite passes 18/18 Python tests and 1/1 CTest. It includes mock protocol coverage for nested autocompile fields, resets, malformed evidence, response IDs, the runtime's lock-free ping response, optional headless frame_perf failure, rolling-window scope, censoring, accounting, and ring ordering/metadata.

I would keep the result explicitly diagnostic, not title correctness, performance, coverage, or production qualification. A locally validated patch exists; I am happy to provide it or open a small focused follow-up if that would be useful. I have not pushed a branch or opened another PR.

mstan:
Thanks @Alexbeav — you're right on the parsing bug, sorry about that.  
Confirmed it against `handle_autocompile_status` and then against a live run. Please double check me again on this. 

Went ahead and pushed a fix. Saw a few non-zeros come back this go around. Enriched the output a bit too.

You were right about `phase_profile` too. It's a rolling window and I'd labelled
it cumulative since boot. It now reads the runtime's own `window_s` and says
plainly that the scope doesn't match the report's window.

**On the rest — yes please, I'd very much like the follow-up.** Fail-closed on
missing/non-numeric/decreasing counters, the dispatch percentages not being a
coverage or time-share measure, top-set censoring in the deltas, the ring
metadata, and schema/tool-hash/atomic output. All fair, and I'd rather have your
version than try to guess on implementation, especially since you've already validated it.
You're always welcomed to open up PRs -- I'm happy to review.

I can try to take a pass myself if you'd like; but I suspect you're probably better equipped. But do let me know, and I can give it a best effort.

Agreed on keeping this explicitly diagnostic and not any kind of correctness or
qualification signal. Worth saying in the tool's own output, not just the PR.

Appreciate you actually running it against a live runtime to help shake this out.

mstan:
Resolved by #133 (merge commit 7e11242), which branched from this PR's head 6a601ef8 and took its slot in the stack.

GitHub auto-resolved this as **merged** rather than closed: #133 was retargeted onto `fix/issue72-overlay-project-root` (#130) and merged there, which put both of this branch's commits (a3b7f47d, 6a601ef8) into this PR's own base — so it closed itself as merged. That is the more accurate outcome than a manual "closed as superseded": the commits landed with attribution intact, they were not discarded. Recording it here so the state is not mistaken for an ordinary merge of this branch on its own.

It went onto #130 rather than `master` deliberately. `review/pr131-hardening-followup` carried e3ecd67b (the BIOS-profile-vs-`--project-root` fix) transitively, so retargeting #133 straight to `master` would have landed a recompiler behaviour change alongside a diagnostics-only tool. #130 stays the single gate for all of it.

What #133 keeps from here: the nested `compile`-object fix and the rolling-window label on `phase_profile`. What it adds: dispatch counters stay raw event counts instead of being rendered as a coverage percentage, fail-closed handling of missing/malformed/reset evidence, explicit censoring labels on the top-N hot sets and the bounded ring, hash-bound atomic JSON artifacts, and 20 protocol tests registered with CTest.

The concrete argument for the swap: on a ring with a clock reset, this branch printed `-10.19 Mcyc/s` / `-0.30 xRT` in section [3] as though it were a measurement. #133 abstains instead. Same failure mode this PR was written to prevent — an empty `phase_hot` read as a bad query — one layer further down.

PR #16, opened 2026-07-12
Improve SmackDown 2 runtime stability, audio, and Vulkan performance

shaneomac1337:
## What changed

- fixes recompiler boot/discovery gaps and overlay-loader hot-path overhead
- adds PSX SPU Gaussian interpolation and reverb support
- improves Vulkan frame delivery with coalesced command submission and reusable staging buffers
- prefers mailbox presentation for frontend-paced tear-free output
- adds Vulkan CRT/composite/Trinitron presentation shaders and launcher selection
- includes Windows framework build and regeneration helpers

## Why

SmackDown 2 previously encountered boot freezes, missing reverb, and Vulkan frame-time spikes caused by repeated command submissions, staging allocations, and competing presentation pacing.

## Impact

The game boots reliably, retains higher-quality PSX audio, and renders through Vulkan at a much steadier cadence while preserving configurable screen simulation.

## Validation

- regenerated the SmackDown 2 static recompilation output
- rebuilt `SmackDown2Recomp.exe` successfully with Vulkan enabled
- exercised Vulkan diagnostic captures at a stable 60 fps

The generic BIOS-only framework link still reports the existing unresolved `psx_game_address_in_text` symbol; the game target itself builds successfully.

PR #133, opened 2026-08-11
tools: harden stall report evidence semantics

Alexbeav:
## Summary

I adapted the locally validated hardening patch onto PR #131 head 6a601ef8. That head already fixes the nested autocompile parsing bug and labels phase_profile as a rolling window, so this follow-up keeps those fixes and limits its diff to the remaining evidence-semantics work.

This change:

- keeps dispatch_native and dispatch_interp_fallback as raw event counts instead of presenting their ratio as coverage or time share;
- fails closed on missing, malformed, mismatched, decreasing, or reset evidence;
- validates normal response IDs while accepting the runtime's exact lock-free ping response;
- records requested and reported rolling-phase scope and labels boundary approximation;
- marks phase_hot and dirty per-PC localization as censored or partial, with aggregate accounting;
- preserves raw bounded-ring entries plus requested/returned counts, capacity, total events, sequence bounds, wrap state, and sampled span;
- keeps ring-based causal language explicitly heuristic;
- writes schema-versioned, tool-hash-bound diagnostic artifacts atomically; and
- adds focused mock-protocol coverage registered with CTest.

## Tests

- Python focused suite: PASS, 20/20 with UTF-8, no bytecode, and warnings as errors.
- Registered CTest stall_report: PASS, 1/1.
- AST parse, 100-column check, diff checks, and clean-worktree check: PASS.

I also ran the broader configured CTest set: 39/44 passed. The five failures are unrelated runtime-tree baseline tests (fps_telemetry_opt_in, mod_load_acceleration, release_zip, sdl3_main_single_include, and vk_present_wait_stage); this follow-up changes no runtime file or input involved in those failures.

## Live validation

I ran the committed tool in an isolated known-title warm-cache sandbox. Both snap --out and a bounded 12-second run --out completed through a transition-stall shape. I then independently reloaded the raw JSON and recomputed the report rather than trusting its rendered text.

The recomputation matched:

- loader deltas: native 20,182; fallback 36,895,109; stale 0;
- dirty deltas: blocks 36,873,812; instructions 103,633,043; native handoffs 0;
- autocompile deltas: runs 0; fails 0; shard failure total 0;
- 13-second requested/reported rolling phase window with boundary warning;
- static hot-set censoring: 64 returned, two before-only and two after-only;
- dirty per-PC accounting: 1,164,646 accounted and 102,468,397 unaccounted;
- ring scope: 95/2,048 returned, 449,318 total events, capacity 16,384, wrapped, sequence 433,057..449,235, span 5,457.107 ms; and
- largest gap: 84.166 ms / 1,055,061 guest cycles / 0.370119x, with at least one top gap inside an exception.

Both artifacts embed the committed stall_report.py SHA-256 f2f74a996dcb3df6e544dcfe07b9dda18215fa975d0fffd910fe817a47452c1a. The runtime was stopped after capture, and the isolated debug listener was closed.

## Scope

This output remains diagnostic evidence. It does not qualify title correctness, performance, coverage, production readiness, or a root cause by itself.

### DIFF
--- a/docs/COMPILING_OVERLAYS.md
+++ b/docs/COMPILING_OVERLAYS.md
@@ -198,6 +198,22 @@ dev config lands under a different tag and the shipped runtime will not read
 it. `--project-root` is what lets you do that without also having to move the
 BIOS profile.
 
+**Is the cache actually being USED?** Shards on disk prove nothing — the loader
+can reject them. `tools/stall_report.py` answers that, plus interpreter
+attribution, in one command against a running title:
+
+```sh
+py -3 tools/stall_report.py --port <debug port> snap            # since boot
+py -3 tools/stall_report.py --port <debug port> run --secs 60   # windowed
+```
+
+It is a passive ring consumer: no arming, no pausing, no stepping. It reads
+cumulative counters (two snapshots → a window by subtraction) and the always-on
+PC-sample ring, so you can attach long after the interesting thing happened.
+`dispatch_native` vs `dispatch_interp_fallback` is the verdict; the
+instructions-per-dispatcher-round-trip ratio tells you whether interpreted code
+is chaining locally or thrashing per-block.
+
 **Verify, don't assume.** Run the preflight and read the result line:
 
 ```sh
--- a/recompiler/CMakeLists.txt
+++ b/recompiler/CMakeLists.txt
@@ -408,6 +408,11 @@ if(BUILD_TESTING)
                     ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_bios_backend_constant_codegen.py
                     --recompiler $<TARGET_FILE:psxrecomp-bios>
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/..)
+        add_test(
+            NAME stall_report
+            COMMAND ${Python3_EXECUTABLE}
+                    ${CMAKE_CURRENT_SOURCE_DIR}/../tools/tests/test_stall_report.py
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/..)
 
         # ---- codegen tests that were written but never registered -----------
         # These existed in recompiler/tests/ with no add_test() entry, so they
--- a/tools/stall_report.py
+++ b/tools/stall_report.py
@@ -20,9 +20,8 @@ Answers, in a single pasteable report:
 PASSIVE RING CONSUMER - this is the whole point. It does not arm a trace, it
 does not pause, and it does not single-step. Every number is either a
 cumulative-since-boot counter (two snapshots -> a window by subtraction) or a
-read of the always-on starvation ring, which has been recording since the
-process started. You can therefore attach at any time, including long after
-the interesting thing happened, and still see it.
+read of the bounded starvation ring. The ring overwrites old entries. Inspect
+its reported span before you attribute an earlier event.
 
 Usage:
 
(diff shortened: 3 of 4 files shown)


mstan:
## Review: approving

Verified independently rather than taking the PR description at face value. Every field this tool now *requires* was checked against the actual format strings in `runtime/src/debug_server.c` on `origin/master` — not against `debug_client.py`, which is not the command inventory.

### Protocol contract — all required fields confirmed present on master

| tool requires | emitter | verdict |
|---|---|---|
| `phase_profile.window` param, `window_s` | `handle_phase_profile` clamps to `PHASE_RING_SECS - 2` | ✅ `PHASE_RING_SECS 64` → `PHASE_RING_MAX_WINDOW = 62` is exactly right |
| all six `*_share` as reals | `%.4f`, all six unconditional | ✅ `require_real` is safe |
| `phase_hot.phase_samples_total`, `.hash_drops`, `top[].addr/.samples` | `handle_phase_hot`, `top` clamped to 64 | ✅ `PHASE_HOT_TOP = 64` matches the server clamp |
| `starv_ring.total`, `entries[].seq/us/cyc/in_exc` | `handle_starv_ring` | ✅ |
| `STARVATION_RING_CAP == 1 << 14` | `runtime/include/starvation_ring.h:30` | ✅ |
| `autocompile_status.compile.degraded` read as an int | `autocompile_status_json` emits `"degraded":%d` | ✅ — would have been a hard fail if it were a JSON bool |
| `configured/state/runs/fails/last_exit/consecutive_fails/shard_*` | same | ✅ |
| `overlay_loader_status.dispatch_native/_interp_fallback/stale_blocked` | `handle_overlay_loader_status` | ✅ |
| `dirty_ram_stats.{blocks,insns}_run`, `per_pc[].pc/insns/hits/entries`, `text_*` | `handle_dirty_ram_stats` | ✅ |
| `kernel_bless.entries/clean/mismatch/native_hits` | `handle_kernel_bless` | ✅ |

The `ping` id-0 exemption is exact, not a fudge: the reply is a hardcoded string in the **IO-thread lock-free fast path** (`io_thread_main`), emitted before dispatch ever sees the request, so it can never carry the caller's id.

### The strict monotonicity check is safe

I was worried `us[i] <= us[i-1]` would abstain from section [3] on a microsecond tie. It can't: PC samples fire on a `psx_pc_sample_throttle >= 1048576` cycle gate — ~31 ms apart at 1x — so consecutive samples cannot share a microsecond outside a genuine clock reset, which is exactly the case that *should* abstain.

### Behavioural verification

Ran the committed tool against a mock server emitting payloads shaped byte-for-byte like the `send_fmt` format strings above (not the PR's own fixtures), in both `snap` and `run` modes:

* Both modes exit 0 and render all six sections.
* `--out` into a non-existent nested path creates it, writes atomically, and leaves **no** `.tmp-<pid>` behind. The embedded `tool.sha256` recomputes to the committed file.
* Ring provenance renders correctly: `returned 300 / requested 2048 / capacity 16384`, wrap detected, seq bounds `449000..449299`.
* Side-by-side against #131 head on a deliberately clock-reset ring: **#131 printed `-10.19 Mcyc/s` / `-0.30 xRT` as if it were a measurement; this PR abstains.** That single diff is the case for the whole change.
* `test_stall_report.py`: 20/20.
* Fresh `cmake -S recompiler -B … -DBUILD_TESTING=ON` configures clean; `ctest -N` lists `#16 stall_report` out of 44; `ctest -R stall_report` passes.

### Two follow-ups (not blockers)

1. **The registration guard does not cover the new directory.** `runtime/check_test_registration.cmake` globs only `runtime/tests/` and `recompiler/tests/`. This PR creates `tools/tests/`, which it registers correctly — but the *next* file dropped there would be an orphan again, silently, which is precisely the class that guard exists to close. Worth adding `tools/tests/test_*.py` to the `file(GLOB _found CONFIGURE_DEPENDS …)` list.

2. **`phot_delta_lines` now intersects the two hot sets**, so an address that *entered* the top-64 during the window — often the thing you opened the tool to find — is excluded from the listing and survives only as a count in the censoring note. Strictly more honest than #131's treat-missing-baseline-as-zero, and the note does disclose it, but the lost signal is worth a bounded "new entry, delta ≤ N" line later.

Neither touches runtime code. Merging.

PR #134, opened 2026-08-11
tests: repair standalone netplay snap-ring harness

Alexbeav:
## Summary

The standalone snap-ring test currently includes `../src/netplay_snap_ring.c` directly, but its documented command also compiles that source separately. That produces duplicate symbols. Compiling the intended single translation unit then fails because the newer VRAM-incremental query functions have no test stubs.

This test-only change:

- removes the duplicate source from the documented command;
- adds zero-return stubs for `boot_state_last_vram_dirty_rows()` and `boot_state_last_vram_incremental()`; and
- enables warnings as errors in the standalone recipe.

No runtime source or behavior changes.

## Verification

Frozen/current upstream base: `2db5f9e7aaa91d948f78e7dab55f219c9ea87123`.

Windows 11, WinLibs GCC 16.1.0:

- repaired standalone snap-ring test: all checks passed;
- neighboring netplay hash-confirm test: all checks passed;
- neighboring netplay input-history test: all checks passed;
- neighboring CAUSE.IP2 test: all checks passed;
- `git diff --check`: passed.

The failure and correction were reproduced in a fresh worktree from upstream master.


TechnicallyComputers:
**Verdict: approve** — small, correct test-only fix. One Linux portability nit on the documented recipe.

[PR #134](https://github.com/mstan/psxrecomp/pull/134) (`6c641db`, one file: `runtime/tests/test_netplay_snap_ring.c`)

### What’s right

1. **Duplicate TU** — The harness `#include`s `../src/netplay_snap_ring.c`, so the old recipe that also listed that `.c` on the gcc line was wrong. Dropping the extra source is the right fix.

2. **Missing stubs** — Confirmed on current tree: single-TU build without the new stubs fails to link on `boot_state_last_vram_dirty_rows` / `boot_state_last_vram_incremental`. Zero-return stubs under `NETPLAY_SNAP_RING_TEST_STUB_BOOT_STATE` match the existing stub style and are enough for this harness (ring store/has/peek + stubbed save/load).

3. **`-Werror`** — Matches neighboring standalone netplay tests; good consistency.

4. **Scope** — No runtime behavior change; diagnosis and fix match the failure mode.

### Nit (non-blocking, but real)

Adding `-Werror` makes a **pre-existing Linux issue** fail the documented recipe:

- With `-std=c11`, `clock_gettime` / `CLOCK_MONOTONIC` in `netplay_snap_ring.c` need a feature macro (e.g. `-D_POSIX_C_SOURCE=200809L`).
- Author verified on **Windows** (WinLibs), where the `#if defined(_WIN32)` QPC path is used, so they wouldn’t see this.
- With `_POSIX_C_SOURCE`, the PR-style build runs and reports **ALL PASS**.

Optional follow-up: add `-D_POSIX_C_SOURCE=200809L` to the recipe (or set it in `netplay_snap_ring.c` before `<time.h>`). Not required to land the duplicate/stub fix if CI is Windows-only or doesn’t run this recipe.

### Summary

Safe merge for the stated problem. I’d approve as-is, or ask for the POSIX define in the same PR if you care about the documented gcc line on Linux.

PR #147, opened 2026-08-16
Add opt-in netplay local viewport extraction

mstan:
## Summary
- add opt-in netplay local viewport extraction for native vertical split-screen titles
- add `[netplay] local_viewport = vertical_split` and optional `local_viewport_aspect` parsing
- detect active vertical split-screen draw areas in the GPU layer and expose that state over TCP
- crop presentation to the local netplay peer's half only when netplay is active and the title opted in

## Validation
- `cmake --build psxrecomp\recompiler\build-codex --config RelWithDebInfo --target recompiler_patch_test --parallel`
- `recompiler_patch_test.exe`
- downstream WipEout validation build passed
- downstream WipEout two-peer TCP validation showed `vertical_split.active=1` and `ws.active=1` after loading savestate slot 1

## Notes
Default behavior remains unchanged. The crop path is inactive unless a game's config opts into local viewport extraction, and the aspect hook is further gated by `local_viewport_aspect`.


mstan:
Follow-up pushed: c18d820 corrects split-viewport projection aspect. The previous result used normal 4:3->16:9 squash for a half-screen source; this now derives GTE/GPU squash from (display_width/2)/display_height -> target peer aspect. WipEout two-peer validation from savestate slot 1 reports vertical_split.active=1, ws.active=1, squash=[1143,2048] (~0.558), x_margin=143. Screenshots: validation_netview_gate/WipEoutNetP{1,2}-split-source-projection.png.

PR #152, opened 2026-08-17
fix(sched): require coherent deferred switch boundaries

Alexbeav:
## Summary

- require a coherent materialized guest boundary before a deferred thread switch
- permit title-neutral dirty-RAM transfer boundaries only when the live and published guest PCs match, allowing KSEG aliases
- add focused regression coverage for accepted and rejected boundary states

## Found in

Metal Gear Solid (Europe), Disc 1 (`SLES-01370`). During the first memory-card scan, a deferred switch could save an idle-loop resume PC together with a live callback register context. Restoring that mixed context prevented the cinematic-to-codec handoff.

## Validation

- Full 23,258-frame canonical PAL route completed with no fatal or unknown dispatch.
- Focused MinGW boundary test passed.
- Focused and six adjacent MSVC tests passed: deferred switching, CAUSE/IP2, cycle boundaries, instruction cache, chain handler, and CD-ROM IRQ mask.
- Independent control: Crash Bandicoot 2 (USA), `SCUS-94154`, completed its qualified clean Cross route through frame 8001 with no fatal or unknown dispatch.
- The implementation contains no title address, serial, region check, or game-specific switch.

## Notes

A broad MSVC build still encounters pre-existing GNU/MSVC portability failures in unchanged source. The changed `interrupts.c` compiles successfully.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -35,6 +35,12 @@ if(BUILD_TESTING)
     add_test(NAME cause_ip2_combinational_test
              COMMAND cause_ip2_combinational_test)
 
+    add_executable(deferred_switch_boundary_test
+        tests/test_deferred_switch_boundary.c)
+    target_include_directories(deferred_switch_boundary_test PRIVATE include)
+    add_test(NAME deferred_switch_boundary_test
+             COMMAND deferred_switch_boundary_test)
+
     add_executable(psx_cyc_batch_test tests/test_psx_cyc_batch.c)
     target_include_directories(psx_cyc_batch_test PRIVATE include)
     target_compile_definitions(psx_cyc_batch_test PRIVATE
--- a/runtime/include/interrupts.h
+++ b/runtime/include/interrupts.h
@@ -9,6 +9,33 @@ extern "C" {
 
 struct CPUState;
 
+/* A deferred cooperative thread switch may snapshot CPUState only after the
+ * guest PC and all architectural registers reach the same committed boundary.
+ *
+ * Site 0 is an ordinary compiled block leader. Site 1 is a dirty-RAM transfer
+ * boundary after the instruction and delay slot retire. Other interpreter
+ * pump sites can still have a hidden call/return obligation.
+ *
+ * A site-1 poll can publish its resume PC before CPUState.pc reflects that
+ * boundary. Require both views to identify the same guest address. KSEG
+ * aliases are equivalent. */
+static inline int psx_deferred_switch_boundary_materialized(
+    int dirty_pump_site,
+    int dispatch_depth,
+    int call_unit_depth,
+    uint32_t cpu_pc,
+    uint32_t resume_pc)
+{
+    uint32_t resume_phys = resume_pc & 0x1FFFFFFFu;
+
+    if (dispatch_depth > 1 || call_unit_depth != 0) return 0;
+    if (resume_phys < 0x00010000u) return 0;
+    if (dirty_pump_site == 0) return 1;
+    if (dirty_pump_site != 1) return 0;
+
+    return ((cpu_pc ^ resume_pc) & 0x1FFFFFFFu) == 0u;
+}
+
 /* IRQ bit positions in I_STAT/I_MASK */
 #define IRQ_VBLANK  0
 #define IRQ_GPU     1
--- a/runtime/src/interrupts.c
+++ b/runtime/src/interrupts.c
@@ -46,6 +46,7 @@
  * static BIOS exception handler is NOT interp code, so we clear it around the
  * handler dispatch (see psx_check_interrupts). */
 extern int g_dirty_interp_active;
+extern int g_call_unit_depth;
 extern uint32_t g_dirty_safe_resume_pc;
 
 /* IRQ-delivery context ring (MMX6 VSync-vs-CD-DMA hunt; dumped via `irqctx_ring`). */
@@ -1231,10 +1232,14 @@ void psx_check_interrupts(CPUState* cpu) {
                 s_defer_switch_target  = 0;
                 s_defer_switch_from    = 0;
                 debug_server_log_thread_event(33, cpu, from_tcb, to_tcb, resume_pc);
-            } else if (g_cosim_dirty_pump_site != 0) {
-                /* Not a real suspend boundary: dirty interpreter pump sites expose
-                 * committed PCs for IRQ timing, but the matching CPUState may not
-                 * be materialized yet. */
+            } else if (!psx_deferred_switch_boundary_materialized(
+                           g_cosim_dirty_pump_site,
+                           g_psx_dispatch_depth,
+                           g_call_unit_depth,
+                           cpu->pc,
+                           resume_pc)) {
+                /* Keep the switch pending until CPUState and the published resume
+                 * PC describe one fully committed guest boundary. */
             } else if (resume_pc != 0u && (resume_pc & 0x3u) == 0u &&
                        psx_is_dispatchable(resume_pc)) {
                 /* Materialized clean boundary: re-save the deferred thread cleanly
--- a/runtime/tests/test_deferred_switch_boundary.c
+++ b/runtime/tests/test_deferred_switch_boundary.c
@@ -1,0 +1,55 @@
+#include "interrupts.h"
+
+#include <stdio.h>
+
+static int failures;
+
+#define CHECK(condition, message) do {                                      \
+    if (!(condition)) {                                                     \
+        fprintf(stderr, "FAIL: %s\n", (message));                          \
+        failures++;                                                         \
+    }                                                                       \
+} while (0)
+
+int main(void)
+{
+    CHECK(psx_deferred_switch_boundary_materialized(
+              0, 1, 0, 0u, 0x80012340u),
+          "an ordinary outermost block leader is materialized");
+    CHECK(psx_deferred_switch_boundary_materialized(
+              1, 1, 0, 0x80012340u, 0x80012340u),
+          "an exact site-1 PC match is materialized");
+    CHECK(psx_deferred_switch_boundary_materialized(
+              1, 1, 0, 0xA0012340u, 0x80012340u),
+          "KSEG aliases identify the same guest PC");
+
+    CHECK(!psx_deferred_switch_boundary_materialized(
+              1, 2, 0, 0x80012340u, 0x80012340u),
+          "a nested dispatch is not materialized");
+    CHECK(!psx_deferred_switch_boundary_materialized(
+              1, 1, 1, 0x80012340u, 0x80012340u),
+          "a live call-unit frame is not materialized");
+    CHECK(!psx_deferred_switch_boundary_materialized(
+              2, 1, 0, 0x80012340u, 0x80012340u),
+          "other dirty interpreter pump sites remain ineligible");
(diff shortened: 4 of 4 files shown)


mstan:
Hiya! Thanks for the submission. I ran a regression suite on this one. This actually regress memory card loading for [Ape Escape](https://github.com/mstan/ApeEscapeRecomp). 

ApeEscapeRecomp_KaDb3TJPYG

Whatever this is changing causes regressed behavior there. We'll need to figure out some sort of discriminator to facilitates both games.

mstan:
## TL;DR

**Ape Escape (SCUS-94423) LOAD GAME wedges permanently on the memory-card check with this branch.** Reproduced deterministically, isolated to the two changed files, and cross-checked. The guard this PR relaxes is the fix for exactly that softlock.

## Repro

Ape Escape, headless, `bios_boot=HLE` so the title menu comes up on its own. Identical guest input at identical **guest frames** on both arms — `UP` at frame 6800 (moves the highlight NEW GAME → LOAD GAME), `CROSS` at frame 6900, then **no further input at all**. Screenshot ladder at +400/+800/+1500/+3000/+6000/+10000/+16000 frames.

| +16000 frames after selecting LOAD GAME | result |
| --- | --- |
| `master` + title pin | memory-card file list — `WLau1 100%`, `WLau2 83.3%`, `File3 83.3%`, `File4 No Data`; selecting a file loads into the Time Station |
| this branch | still `Checking··· / Do not remove MEMORY CARD`. Never advances. |

That is ~4.5 minutes of guest time on a screen that takes ~1200 frames to clear on `master`. It is a hard wedge, not a slowdown. Also confirmed hands-on with a real controller, not just headless.

## Scheduler telemetry

Sampled continuously from the always-on thread-event ring across the whole run (kind 31 = deferral armed, kind 32 = deferral honored):

| kind | `master` | this branch |
| --- | --- | --- |
| 31 (defer armed) | 7878 | 23 |
| 32 (defer honored) | 7878 | 23 |
| 20 / 3 / 8 | 7878 each | 15745 each |
| 2 | 15756 | 31490 |

So the branch does what it intends — it stops deferring at these boundaries — and the card manager then spins instead of completing. The counts are stable across repeats (a second pair of runs gave 7872 / 42).

## Isolation

- The two executables differ **only** in `runtime/src/interrupts.c` and `runtime/include/interrupts.h` (directory-diffed the whole framework tree; the patch was applied on top of the title's own pinned framework via a `PSXRECOMP_ROOT` override, so no submodule pin moved and no codegen changed).
- Each arm ran from its own sandbox directory with its own copy of the same memory card.
- **Cross-swap control:** `master` exe + *this branch's* card copy → reaches the file list. This branch's exe + *master's* card copy → wedges. The failure follows the executable, not the card, the cache, or the sandbox.
- 4 runs per arm, unanimous.

## Root cause

The guard being relaxed —

```c
} else if (g_cosim_dirty_pump_site != 0) {
```

— is `0cb2a77b`, *"fix(irq): gate deferred thread-switch off dirty-pump and low-kernel sites"*. That commit exists **specifically for Ape Escape memcard save/load**, and was validated hands-on at the time.

Its stated hazard is:

> the deferred in-exception cooperative thread-switch could be honored at a boundary where a candidate resume PC was committed but **the live CPUState (stack/GPRs)** still described the previous local frame

`psx_deferred_switch_boundary_materialized()` substitutes a **PC** coherence test for that. `cpu->pc == resume_pc` does not establish GPR/stack coherence — it establishes only that two views agree on the program counter, which at a site-1 boundary is true almost by construction: `dirty_ram_interp.c:3049` calls the pump with `target = cpu->pc`, so the committed PC *is* `cpu->pc` at entry, and nothing between there and the honor site mutates it while a deferral is pending. In practice the new site-1 predicate is close to unconditional.

`dispatch_depth`/`call_unit_depth` don't cover the gap either: an ordinary site-1 boundary is reached at dispatch depth 1 and call-unit depth 0, both of which the predicate accepts.

## Two further issues, independent of the wedge

**1. The honor path strands the dirty-pump latches.** It ends in `longjmp(g_scheduler_jmpbuf, 1)`. Reaching it from site 1 means longjmping out of `dirty_ram_pump_boundary()`, which skips its own restores:

```c
/* runtime/src/dirty_ram_interp.c:864-865 */
g_dirty_safe_resume_pc  = prev_safe;
g_cosim_dirty_pump_site = prev_site;
```

The scheduler-escape landing at `traps.c:812` resets `g_psx_dispatch_depth`, `g_psx_call_bail`, `psx_in_device_service`, `g_call_unit_depth` and the cycle-batch state — but **not** those two. A stale non-zero `g_dirty_safe_resume_pc` is then consumed as:

- `real_pc`, i.e. the EPC for the **next** IRQ delivery, under the newly-scheduled thread (`interrupts.c:1468`), and
- `s_last_interrupt_check_pc`, which feeds savestate / netplay / rewind / idle-skip (`interrupts.c:1098`, `1129`, `1152`).

`g_cosim_dirty_pump_site` stuck at `1` also feeds this PR's own predicate on the next deferral. Fix shape: clear both next to the existing `g_dirty_interp_active = 0;` in the honor path, or better, add them to the escape landing alongside the other invariants.

**2. No load-delay flush at site 1.** `dirty_ram_pump_boundary` at site 1 doesn't call `dirty_ram_ld_delay_flush()` (`dirty_ram_interp.c:336`), whose contract is *"flushed on interpreter exit and before exception delivery"*. `psx_sched_save_context` → `psx_save_context_to_tcb` (`traps.c:314`) snapshots `cpu->gpr[]` directly, so an owed writeback is neither in the saved TCB nor discarded — and the interpreter's pending-load static survives the longjmp, to be retired into whatever thread runs next. Narrow in practice (the deferral only arms for BIOS-ROM/kernel-RAM PCs, and this PR's own `resume_phys < 0x10000` rule excludes the kernel-RAM half), but it's the same class as the `cardfasttrack.s` bug the deferred-load model was written for.

**3. Comment drift.** The block comment above the honor site still says *"honor it only once execution reaches a normal interrupt poll with site 0"*, which the change makes untrue. And in the new header, "Site 0 is an ordinary compiled block leader" is imprecise — site 0 is the *default* value, and the in-interpreter MTC0 software-interrupt pump (`dirty_ram_interp.c:3016`) also runs under it.

## Why the other titles came back clean — and why that isn't reassurance

I ran the full frame-fingerprint A/B (per-frame write / PC / MMIO / SP hashes and write counts) across Tomba 1, Tomba 2 and MMX6, 20,000 guest frames per arm, two index-matched runs each:

- **Tomba 1** — byte-identical on every compared frame. ~116k thread events sampled; **kinds 30–33 never fire.**
- **MMX6** — byte-identical. ~77k events; kinds 30–33 never fire.
- **Tomba 2** — byte-identical at boot depth; deeper soaks inconclusive because the *within-arm* control also 

**2. No load-delay flush at site 1.** `dirty_ram_pump_boundary` at site 1 doesn't call `dirty_ram_ld_delay_flush()` (`dirty_ram_interp.c:336`), whose contract is *"flushed on interpreter exit and before exception delivery"*. `psx_sched_save_context` → `psx_save_context_to_tcb` (`traps.c:314`) snapshots `cpu->gpr[]` directly, so an owed writeback is neither in the saved TCB nor discarded — and the interpreter's pending-load static survives the longjmp, to be retired into whatever thread runs next. Narrow in practice (the deferral only arms for BIOS-ROM/kernel-RAM PCs, and this PR's own `resume_phys < 0x10000` rule excludes the kernel-RAM half), but it's the same class as the `cardfasttrack.s` bug the deferred-load model was written for.

**3. Comment drift.** The block comment above the honor site still says *"honor it only once execution reaches a normal interrupt poll with site 0"*, which the change makes untrue. And in the new header, "Site 0 is an ordinary compiled block leader" is imprecise — site 0 is the *default* value, and the in-interpreter MTC0 software-interrupt pump (`dirty_ram_interp.c:3016`) also runs under it.

## Why the other titles came back clean — and why that isn't reassurance

I ran the full frame-fingerprint A/B (per-frame write / PC / MMIO / SP hashes and write counts) across Tomba 1, Tomba 2 and MMX6, 20,000 guest frames per arm, two index-matched runs each:

- **Tomba 1** — byte-identical on every compared frame. ~116k thread events sampled; **kinds 30–33 never fire.**
- **MMX6** — byte-identical. ~77k events; kinds 30–33 never fire.
- **Tomba 2** — byte-identical at boot depth; deeper soaks inconclusive because the *within-arm* control also diverges (run-to-run persisted-state nondeterminism), not because of this change.
- Their memory cards were then validated hands-on on this branch — all three load saves fine.

The reason those are clean is that `s_defer_switch_pending` is never set during boot/attract, so the modified branch is never evaluated at all. **Ape's boot and attract were byte-identical too** — 20,000 frames, zero deferred-switch events. The regression is completely invisible until the game is driven into the memory-card flow. So "no diff on a long route" is not evidence for this change; the deferred-switch counters have to be non-zero for the run to have tested anything.

Worth noting your MGS route and the Crash 2 control may be in the same position — it'd be worth checking whether kinds 31/32 are non-zero on the Crash run before treating it as a negative control.

## Where this leaves us

I'd like to land a fix for the MGS handoff, but it needs a boundary predicate that establishes what `0cb2a77b` was actually protecting — **register/stack coherence** — rather than PC coherence, and it has to keep Ape Escape's LOAD GAME working. Happy to run the Ape harness against a revised version; the check is:

> title menu → `UP` → `X`, no further input, must reach the memory-card file list within ~1500 frames.

Alexbeav:
Thanks for the deep review — and for the Ape Escape repro in particular. You're
right on all three counts, and the regression is confirmed on our side too.
Point-by-point:

**1. The PC-coherence predicate is nearly unconditional at site 1.** Confirmed.
`resume_pc` is latched from the same check call that set `cpu->pc`, so the two
views coincide by construction and the XOR test gates nothing. We've de-drifted
the comment so the code no longer claims a safety property it doesn't have, and
we agree the real fix is a register/stack-coherence predicate rather than a PC
comparison. We're designing that now; validation plan below.

**2. Stranded latches across the honor longjmp.** We had independently found
and cleared `g_dirty_safe_resume_pc` (same failure signature you describe: a
frames-old pump PC restored into the wrong thread → teleport → AdEL), but you
caught one we missed: `g_cosim_dirty_pump_site` is stranded by the same
longjmp class. Both are now cleared in both honor paths (kind-34 immediate and
kind-32 materialized-boundary) and, as a backstop, at the scheduler landing
alongside the other invariant resets.

**3. Missing load-delay flush at site 1.** Also confirmed and fixed: the pump
can honor a switch via longjmp and never return to the interpreter frame, so an
owed load writeback would retire into the switched-to thread's register file.
We now drain the pipeline (`dirty_ram_ld_delay_flush`) before the site-1
boundary call, matching the existing exception-entry flush.

**Validation.** We'll adopt your Ape acceptance test as a mandatory gate for
this whole family (title menu → UP → X → memory-card file list within ~1500
frames), alongside our own deterministic MGS route, and we'll treat negative
controls as valid only when the kind-31/32 counters are non-zero — your
7878-vs-23 telemetry made it clear how easy it is to "pass" a test the code
path never ran. If the offer of the Ape harness stands, we'd gladly take it.

We'll withdraw the current revision and refresh the series with the fixes
above once they've passed both gates.

mstan:
Sounds like a plan. Eagerly awaiting that follow up. Closing for now. We can re-open this one or start a new one once you have it ready to go. 

Take care!

PR #162, opened 2026-08-19
Forward a pending load into an LWL/LWR merge

Unchiga:
﻿LWL/LWR take their destination register as both the merge base and the destination, and they read that base late enough to receive the forwarded result of an immediately preceding load. They are therefore not subject to the MIPS-I load delay. After

```asm
lw   $t0, 12($a3)
lwr  $t0, 10($a3)
```

the LWR merges into the word the LW just fetched, not the stale `$t0`.

The load-delay pair emitter defers a load's writeback into a `psx_ldd_<addr>` temporary, so an LWL/LWR successor was merging into `cpu->gpr[rt]`, which still held the pre-load value.

## The change

`translate_lwl` / `translate_lwr` now take their merge operand from `lwlr_merge_operand(rt)` instead of naming the register directly. When the block emitter sees that the successor half of a deferred load pair is an LWL/LWR writing the same register, it points that operand at the deferred temporary and suppresses the ordinary writeback, because the merge has already consumed it.

Only the consumer side needs this: LWL/LWR as the *producer* never defer, so their writeback is already immediate.

## Verification

Regenerated a title that exercises the pair: **16 forwarding sites**, and every emitted merge operand identical to the reference output built with the fix in place.

Two files, +51/-3, sitting directly on `master` with no other changes.

### DIFF
--- a/recompiler/CMakeLists.txt
+++ b/recompiler/CMakeLists.txt
@@ -404,6 +404,12 @@ if(BUILD_TESTING)
             COMMAND ${Python3_EXECUTABLE}
                     ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_overlay_shim_compile.py)
         add_test(
+            NAME lwlr_load_delay_forward
+            COMMAND ${Python3_EXECUTABLE}
+                    ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_lwlr_load_delay_forward.py
+                    --recompiler $<TARGET_FILE:psxrecomp-game>
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/..)
+        add_test(
             NAME dirty_text_continuation_guards
             COMMAND ${Python3_EXECUTABLE}
                     ${CMAKE_CURRENT_SOURCE_DIR}/../runtime/tests/test_dirty_text_continuation_guards.py)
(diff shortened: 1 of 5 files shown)


mstan:
Added a second commit to complete this. The same defect was in `full_function_emitter.cpp` (BIOS / resident functions), which this PR did not touch: a `lw rt` followed by `lwl/lwr rt` was deferred and then *discarded* — since LWL/LWR writes `rt`, the flush-vs-discard rule emitted `(void)psx_ldd_<addr>` — so the fetched word vanished and the merge combined the stale register. It now forwards the same way the CFG emitter does, and throws rather than emitting a stale merge if the translator's emitted shape ever changes.

Also added `test_lwlr_load_delay_forward.py`, registered in CMake. It fails on master (merge operand `cpu->gpr[8]`, plus a discard) and passes here, so the bug is pinned rather than just described.

Regen check before merging: Ape Escape, Mega Man X6 and the SCPH1001 BIOS all come out byte-identical before and after — no title has a dependent load-delay pair today, since compilers schedule around the delay. So this is latent-path hardening with no output change to any shipping title. It is an emitter change, so the cg tag bumps and overlay shards rebuild on first run.

Thanks for this one — it closes a real native/interp divergence, since the interpreter already forwards here.

PR #169, opened 2026-08-20
Overlay DLL linking, variant-scoped captures, PGXP shared-edge correctness, and guarded LTO

tetrisgm:
Framework fixes and instrumentation surfaced while bringing up a title that stresses overlays, HD texture replacement and PGXP at a doubled CRTC rate. Everything here is general; the title's own configuration stays in its game repository.

Full per-change audit, with root causes at `file:line`, covering tests, risk review and a suggested split: **`docs/UPSTREAM_2026-08-20.md`**.

## The headline bug

**Overlay DLLs never linked, so overlays silently fell back to per-instruction interpretation.**

`compile_overlays.py` injected `overlay_dispatch_preamble.c.inc` after the *syntactically last* `#include`. A conditional i-cache include moved that anchor inside an `#else`, and DLL builds define `PSX_OVERLAY_DLL_BUILD` — so the callback definitions preprocessed away and every shard failed to link on ~20 runtime symbols. The runtime's own exit report showed `disp_native = 0` alongside millions of interpreted dispatches, with `autocompile_degraded = 1` naming the cause.

This affects any title using overlays, and it fails *quietly* — the game still runs, just far slower. The shim is now anchored to the unconditional `#include "psx_runtime.h"` and raises if that anchor is ever absent, with a regression test asserting the shim precedes the DLL conditional.

## Correctness

- **Variant-scoped overlay captures.** Stale PC bitmaps survived when executed code later became data, so one variant's PC set could be compiled against another variant's bytes. `overlay_watch_note_write` now retires the whole page's execution/dispatch epoch when any executed word is overwritten; static demand is keyed by `(load_addr, size, bytes)`.
- **VRAM→VRAM copies were untrackable for texture replacement.** They only invalidated the destination and never registered it as a candidate, so textures arriving by copy could never be dumped *or* replaced. `gr_copy_rect` now calls `tex_pack_on_copy` with the same key semantics as CPU uploads.
- **PGXP shared-edge seams.** A per-primitive all-or-nothing gate cannot stop two adjacent primitives disagreeing about a shared endpoint — the edge gets submitted at two sub-pixel positions, which is the thin-seam artifact. Adds a frame-stamped canonical endpoint cache keyed by final integer screen position: first use fixes the coordinates, every later occurrence reuses them.
- **PGXP provenance.** `PGXPValue` now carries per-component X/Y/Z projection ids, so a value assembled from two different projections cannot arm correction. Destructive `AND`/`XOR`/`NOR`/`SLT`, link and COP0 writes invalidate shadows; the SXYP shadow FIFO shifts with guest SXY.
- **Precise NCLIP no longer replaces guest MAC0.** It was allowed to flip the sign via a cancellation-prone float determinant, which can change face visibility. It is now telemetry only, so culling stays bit-exact with the integer GTE.
- **Removed two counters** that were being cited as correctness evidence but were never incremented anywhere.

## Performance

- **`PSX_ENABLE_LTO`, plus a `noinline` guard on the shared cycle wrappers.** This matters: *unguarded* ThinLTO clones hot interlock bodies and **regresses** the hot path substantially. Guarded, it becomes a solid win. Adds `runtime/tests/bench_psx_runtime_hotpath.c`, which verifies variant equivalence by digest, so codegen changes are measurable rather than argued about.
- **Cycle accounting** now selects specialized arity-0/1/2/3 helpers at generation time instead of a generic dependency-mask loop, and main-RAM loads combine base/wait/completion into one exact charge. Exactness is covered by pipeline-state, load-state, deadline-crossing and cross-device causality tests.
- **HD texture invalidation** replaced a linear scan of up to 8192 upload records per VRAM write with a tile-refcount index.
- **Translation dispatch** gated behind a monotonic mutation epoch instead of a callback on every dispatch; `card_data_writes_check` no longer runs on every RAM write.

## Capability

- **`fixed_outer_aspect`** (default off): the outer presentation aspect becomes a function of user configuration only. Scene classification chooses *content* transforms, never the outer viewport, so menus and FMV stay centred and undistorted inside a wide canvas instead of switching the viewport per frame.

## Verification, and what is not verified

Each item ships with unit or structural tests, and `docs/UPSTREAM_2026-08-20.md` deliberately marks claims **UNKNOWN** where the committed delta contains no result transcript. Measured numbers from the originating session are recorded in that document's appendix as single-machine observations, not a portable benchmark.

**This body of work has not been validated end-to-end against a fixed reproducible workload.** The timing-exactness and PGXP changes are the highest-risk areas and deserve the closest review — the audit's RISK section names the specific invariants each must preserve.

This is a large PR (126 commits). The audit proposes an ordered split if you would rather take it in pieces; I am happy to break it up along those lines.


tetrisgm:
Closing this monolithic PR now that its independently reviewable pieces are submitted as PRs #214–#222. The branch remains available; please review and merge the smaller PRs individually.

PR #179, opened 2026-08-23
sio: gate Ape Escape card IRQ unstick behind config

mstan:
Split from original PR #169 by tetrisgm; original PR intentionally left open.\n\nThis PR extracts only the Ape Escape memory-card IRQ unstick gating change. The pump is now controlled by [runtime] ape_card_unstick instead of being implicitly active for all titles.\n\nValidation run locally:\n- Built psxrecomp-toml\n- Configured runtime with Retcomm Clang/Ninja using PSXRECOMP_ALLOW_NO_BIOS=ON and PSX_RECOMP_UI=OFF\n- Built sio_dualshock_rumble_test\n\nNotes:\n- sio_dualshock_rumble_test is disabled in the upstream CTest registration, so this PR validates compile coverage only for the SIO test target.\n- Runtime configure emitted the existing stale generated BIOS warning; not caused by this branch.

### DIFF
--- a/recompiler/src/config_loader.cpp
+++ b/recompiler/src/config_loader.cpp
@@ -448,6 +448,9 @@ static RuntimeConfig parse_runtime_block(const toml::value& cfg, const fs::path&
     if (runtime.contains("fast_boot")) {
         rt.fast_boot = toml::find<bool>(runtime, "fast_boot");
     }
+    if (runtime.contains("ape_card_unstick")) {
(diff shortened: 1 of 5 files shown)


mstan:
Rejecting this split as an architecture direction: Ape Escape/card correctness cannot be handled through a title-specific runtime config/env gate. Any SIO/card IRQ fix needs to be generic and preserve Ape Escape plus the broader regression suite without requiring per-game correctness opt-ins. Keeping the original PR open; this split can remain historical/diagnostic context only.

PR #181, opened 2026-08-23
mods: add default-off Skip FMVs package

mstan:
Split from original PR #169 by tetrisgm; original PR intentionally left open.\n\nThis PR extracts only the built-in Skip FMVs mod package and runtime plugin. It deliberately keeps the feature default-off and does not include the broader global launcher exposure/default policy from #169.\n\nIncluded:\n- mods/builtin/packages/psx.enhancement.skip-fmv/1.0.0/manifest.toml\n- 
untime/src/mod_builtin_skip_fmv.c\n- runtime source-list registration\n\nValidation run locally:\n- python runtime/tests/test_mod_owned_display_controls.py\n- Configured runtime with Retcomm Clang/Ninja using PSXRECOMP_ALLOW_NO_BIOS=ON and PSX_RECOMP_UI=OFF\n- Built psx-runtime\n\nNotes:\n- I corrected the extracted manifest description to match default_enabled = false; PR #169 had intermediate default-on wording that did not match its final extracted policy.\n- 
untime/tests/test_mod_load_acceleration.py currently fails against this master text for an unrelated fast-loading ownership assertion, so it was not used as validation for this Skip FMV-only branch.

### DIFF
--- a/mods/builtin/packages/psx.enhancement.skip-fmv/1.0.0/manifest.toml
+++ b/mods/builtin/packages/psx.enhancement.skip-fmv/1.0.0/manifest.toml
@@ -1,0 +1,24 @@
+format_version = 5
+id = "psx.enhancement.skip-fmv"
+version = "1.0.0"
+name = "Skip FMVs"
+author = "psxrecomp"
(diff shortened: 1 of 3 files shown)


mstan:
Rejected on policy/UX grounds. Even though this package is default-off, it targets game_id = *, which means every title would show a Skip FMVs option whether or not the game has FMVs or whether the author wants that feature exposed. Skip FMVs should be author-specified per title: keep any reusable framework service/plugin capability if useful, but do not add a global built-in package that appears for all games by default.

PR #183, opened 2026-08-23
mods: enable Fast Loading by default

mstan:
Split from original PR #169 by tetrisgm; original PR intentionally left open.

This is intentionally isolated as a policy-only PR. It globally changes the built-in Fast Loading feature from default-off to default-on for game_id='*'.

Validation:
- git diff --check

Risk / review note:
- This affects every title using the built-in mod catalog, so it should be accepted or rejected independently from the correctness and infrastructure splits.
- Requires heavier title regression if accepted: Mega Man X6, Tomba 1, Tomba 2, Ape Escape, and WipEout scenarios with Fast Loading default-on and explicitly disabled.

### DIFF
--- a/mods/builtin/packages/psx.enhancement.fast-loading/1.0.0/manifest.toml
+++ b/mods/builtin/packages/psx.enhancement.fast-loading/1.0.0/manifest.toml
@@ -15,9 +15,9 @@ game_id = "*"
 [[feature]]
 id = "fast-loading"
 name = "Fast Loading (host pacing)"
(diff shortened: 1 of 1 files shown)


mstan:
Rejected after validation: MMX6 still reproduces the known WARNING-screen pacing regression when Fast Loading is enabled by default. This is the same underlying turbo-load path, bounded to 4x by default but not a detector fix, so it cannot be default-on under the no-regressions policy. Fast Loading can remain available as an optional/default-off mod until the detector/pacing issue is fixed.

mstan:
Additional validation detail: the rejection is specifically because Mega Man X6 still exhibits the known WARNING load-screen pacing problem with Fast Loading enabled by default. In the room/transition where the large WARNING text reveals over gameplay, the 'WAR' portion appears much faster than intended. That means the current detector still classifies visible/non-load presentation as a load-acceleration window, so default-on would preserve bad behavior for MMX6 even though the optional feature is still valuable.

PR #190, opened 2026-08-23
presentation: add configurable GL bezel artwork

mstan:
Split from original PR #169 by tetrisgm; original PR intentionally left open.

Scope:
- Add [video] bezel to game config parsing.
- Load a configured image after the GL context exists.
- Draw the image behind the 15-bit FBO present when letterbox or pillarbox margins are visible.

Kept out of this split:
- Later tiled/team/random bezel refinements, which are dependent on this foundation and more title-specific.
- FMV reconstruction/filter changes, already split separately.
- Timing, texture-pack, PGXP, and high-RAM overlay changes.

Validation:
- git diff --check origin/master...HEAD
- cmake --build build\\split-runtime --target psx-runtime
- cmake --build build\\split-recompiler --target psxrecomp-toml

Notes:
- Manually resolved the first bezel commit against current master and excluded unrelated HD texture, edge-fill, and PGXP context from the original branch.


mstan:
Updated this split after validation/discussion: the bezel renderer primitive remains, but the public surface is now a disabled-by-default Mods feature instead of a raw [video] bezel config key.\n\nCurrent shape:\n- Default behavior is unchanged: no enabled mod means black letterbox/pillarbox margins.\n- Builtin package: psx.presentation.bezel / feature bezel, default_enabled=false.\n- Trusted plugin: psx.bezel resolves package-local bezel.png and calls the runtime bezel renderer hook.\n- Custom/title packages can provide alternate bezel.png data using the same trusted plugin; archives still cannot load native code.\n\nValidation:\n- Tomba preview confirmed bezel rendering works once artwork resolves.\n- python runtime/tests/test_mod_bezel.py\n- git diff --check\n- Rebuilt Tomba runtime against this branch; new mod_builtin_bezel.c/main/mod_runtime/config_loader objects compiled and linked.\n\nNote: local Tomba preview build used copied generated BIOS C and still reports the existing stale BIOS fingerprint warning; that was preview-only and not part of this PR.

PR #191, opened 2026-08-23
SteamOS/Steam Deck self-build: wire pack sysroot into CMake, surface CLI failure lines, toolchain-stamp hardening

TechnicallyComputers:
## Problem

Steam Deck users on stock SteamOS hit **"psxrecomp rebuild failed (exit 1)"** in the setup wizard with no further detail ([mstan/recomp-ui#34](https://github.com/mstan/recomp-ui/issues/34), Tomba 0.12.0-alpha and Klonoa).

Root cause (reproduced in a bwrap stock-SteamOS simulation — tmpfs over `/usr/include`, GL dev symlinks removed — with the shipped cmake-clang-v1 v1.0.14 pack): the pack bundles a jammy build sysroot (glibc, libstdc++, GL headers, legacy `libGL.so` stub), but only `clang.cfg` uses it. CMake's `find_*` still searches the host, so recomp-ui's `find_package(OpenGL REQUIRED)` dies at configure with `Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR)` — and the wizard swallowed everything except the exit code.

## Changes

**1. Wire the pack sysroot into CMake** (`psxrecomp_cli.py`)
- New `_pack_sysroot_cmake_args()`: when the active pack carries `sysroot/usr/include` on Linux, both the game configure and the emitters configure get `-DCMAKE_SYSROOT=<pack>/sysroot -DOpenGL_GL_PREFERENCE=LEGACY`.
- `LEGACY` is load-bearing: the sysroot ships legacy `libGL.so` + `GL/gl.h` but no GLVND dev files (`glx.h`/`libOpenGL.so`/`libGLX.so`), so under GLVND preference `OpenGL::GL` is never created even with the sysroot.
- Skipped when the caller already passes `CMAKE_SYSROOT`, on non-Linux, or with a system clang.

**2. Surface the actual failure in the wizard** (`host/psxrecomp_codegen_host.c`)
- `run_cli_posix`/`run_cli_win` route the child's stderr into the progress pipe and keep a last-lines capture (`CliTail`). On failure, `err_msg` now carries the most recent error-looking line — JSON progress rows contribute their `message`, raw rows (compiler output, Python tracebacks) contribute as-is:
  `psxrecomp rebuild failed (exit 1): CMake Error: Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR)`
- Exit-3 disc-verify message unchanged.

**3. Toolchain hardening ported from feat/rbengine**
- `_toolchain_stamp` + wipe-on-mismatch in `ensure_emitters`: an upgrade behind the stable `latest` symlink (e.g. v1.0.10 → v1.0.14, which changed the glibc objects were compiled against) never trips CMake compiler-change detection and old objects fail to link with `undefined symbol: __isoc23_strtoul`. The stamp resolves the symlink to the concrete pack version; a changed or missing stamp over an existing `CMakeCache.txt` wipes the tree once.
- `toolchain_pack.py`: lazy `urllib` imports so the module still loads when an ambient `PYTHONHOME` breaks `_socket`.

## Verification

- Deck simulation (bwrap, no `/usr/include`, no GL dev symlinks, real v1.0.14 pack): unpatched `_cmake_configure` reproduces the exact Deck failure; patched path configures, builds, and the binary runs.
- `_pack_sysroot_cmake_args` unit-tested (pack clang / already-set / no clang / system clang).
- `CliTail` exercised against real compact-JSON rows (`sdk_progress` format), an `event:error` row, raw stderr tracebacks, success, and exit-3.
- Host file syntax-checked on both Linux (clang) and Windows (mingw) branches.
- Stamp: differs across cached v1.0.10/v1.0.14 packs, resolves `latest` to the concrete version; full `ensure_emitters` build writes it; tampered and missing stamps each trigger the wipe + clean rebuild.

**4. Default toolchain min-version floor 1.0.14** (`tools/toolchain_pack.py` + host)
- The cache was accepted forever once any cmake ran, so a host that cached a pre-sysroot pack (≤1.0.13) never re-downloaded. Both resolvers now floor at 1.0.14 (first pack with the Linux build sysroot). `RETCOMM_TOOLCHAIN_MIN_VERSION` overrides in either direction; `0`/`off`/`none` disables. Verified against the real cached v1.0.10/v1.0.14 packs on both the Python and C sides, including an isolated stale-only cache no longer resolving.

## Companion PR

- [TechnicallyComputers/retcomm-toolchains#2](https://github.com/TechnicallyComputers/retcomm-toolchains/pull/2) — sysroot GLVND completion (`libglx-dev`/`libopengl0`/`libopengl-dev`, pack 1.0.15) so FindOpenGL's default GLVND path also resolves from the pack; the LEGACY preference here keeps 1.0.14 packs working either way.

## Not in this PR

- Tomba/Klonoa releases must be re-cut with a new framework pin for this to reach users.

https://claude.ai/code/session_01Vc7zPEnfUFMhMtJzUEruyb

### DIFF
--- a/host/psxrecomp_codegen_host.c
+++ b/host/psxrecomp_codegen_host.c
@@ -1496,6 +1496,55 @@ static void handle_progress_line(const char* line,
     }
 }
 
+/* Last-lines capture so a failed CLI run can say WHY in the wizard, which has
+ * no console: the child's stderr shares the progress pipe, JSON progress rows
+ * contribute their "message", raw rows (compiler/traceback text) contribute
+ * as-is, and the most recent error-looking line is appended to err_msg. This
+ * is what turns "psxrecomp rebuild failed (exit 1)" into "… failed (exit 1):
+ * Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR)". */
+typedef struct {
+    char last[480];
+    char last_err[480];
+} CliTail;
+
+static int cli_tail_line_is_error(const char* s) {
+    return strstr(s, "rror") != NULL || strstr(s, "ailed") != NULL ||
+           strstr(s, "FAILED") != NULL || strstr(s, "Traceback") != NULL ||
+           strstr(s, "fatal") != NULL || strstr(s, "Fatal") != NULL;
+}
+
+static void cli_tail_note(CliTail* t, const char* line) {
+    char msg[480];
+    const char* rec = line;
+    int is_error_event = 0;
+    if (line[0] == '{') {
+        /* sdk_progress emits compact JSON: {"event":"error","message":…}. */
+        is_error_event = strstr(line, "\"event\":\"error\"") != NULL;
+        if (!json_get_string(line, "message", msg, sizeof(msg)))
+            return; /* structured row without text (e.g. result) */
+        rec = msg;
+    }
+    if (!rec[0])
+        return;
+    snprintf(t->last, sizeof(t->last), "%s", rec);
+    if (is_error_event || cli_tail_line_is_error(rec))
+        snprintf(t->last_err, sizeof(t->last_err), "%s", rec);
+}
+
+static void cli_fail_msg(char* err_msg, size_t err_cap, const char* fail_label,
+                         long code, const CliTail* t) {
+    const char* why = t->last_err[0] ? t->last_err : t->last;
+    if (code == 3) {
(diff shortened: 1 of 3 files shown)


TechnicallyComputers:
Closing per plan change: this work will land via the feat/rbengine-bak line instead of a separate master branch. The branch fix/steamdeck-selfbuild stays intact if we want to reopen.

PR #195, opened 2026-08-23
recompiler: analysis DB, widescreen scan, config schema (stack 1/4)

TechnicallyComputers:
Slice 1 of 4 — the reviewable split of `feat/rbengine-bak` into stacked PRs. Merge order: this PR first; each later slice bases on the previous. **Slice 4's tree is byte-identical to `feat/rbengine-bak`** (verified via `git rev-parse HEAD^{tree}`), so merging the whole stack lands exactly the integration branch, no more, no less.

Recompiler-domain unit (18 files, no runtime code):

- **analysis_db / analysis_export / main_analyze** — queryable function-discovery database + export pipeline (`docs/FUNCTION_DISCOVERY.md`)
- **bios_call_names** — symbolize BIOS service calls in analysis output
- **widescreen_scan** — static scan for cull/aspect candidate sites, consumed by the widescreen runtime in slice 4
- **config_loader** — `game.toml` schema for the whole stack (pgxp, tex_pack, widescreen, link/8mb), landed first so later slices compile against final config structs; `docs/config_schema.md` updated
- **code_generator / full_function_emitter / main_psx / ps1_exe_parser** — emitter support the runtime slices rely on

Verified: recompiler builds clean; standalone `runtime/` (`psx-runtime`) builds clean against this slice (master runtime + new config schema).

Stack: **1/4 (this)** → [2/4 tools] → [3/4 project_studio] → [4/4 runtime]

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016mwbhSkoL66o8BSBuP6DiY

PR #196, opened 2026-08-23
tools: DuckStation oracle, GPU frame toolkit -- CD-ROM Controller Patch (stack 2/4)

TechnicallyComputers:
Slice 2 of 4 of the `feat/rbengine-bak` stacked split (base: #195). Standalone Python/container tooling — no runtime or recompiler code:

- **duckstation_oracle.py + tools/duckstation/** (Containerfile, pin.json) — containerized DuckStation as an additional comparison oracle, Linux build support
- **gpu_frame_capture / gpu_frame_diff / gpu_frame_layers / psx_gpu_frame / gpu_parity** — frame-level GPU diagnostics against the oracle (`docs/GPU_FRAME_TOOLS.md`)
- **coverage_vault, debug_client, compile_overlays, toolchain_pack, package_setup_host** — workflow updates used by the later slices' features
- `tools/tests/` for the oracle and frame pipeline

Verified: 25 oracle tests pass; all tools byte-compile. (`test_gpu_frame.py` needs numpy, which is currently broken at the host level on the assembly machine — glibc skew, unrelated to this diff.)

Stack: [1/4 recompiler #195] → **2/4 (this)** → [3/4 project_studio] → [4/4 runtime]

https://claude.ai/code/session_016mwbhSkoL66o8BSBuP6DiY

### DIFF
--- a/docs/GPU_FRAME_TOOLS.md
+++ b/docs/GPU_FRAME_TOOLS.md
@@ -1,0 +1,302 @@
+# GP0 frame tools — who drew this pixel?
+
+A set of tools that answer one question about a rendering bug: **which guest
+function issued this primitive, and with which blend mode?**
+
+```
+tools/psx_gpu_frame.py       transport + GP0 decode + attribution (the library)
+tools/gpu_frame_capture.py   capture a frame from a running game
+tools/gpu_frame_layers.py    render one image per guest function
+tools/gpu_frame_diff.py      diff a good frame against a bad one
+tools/gpu_parity.py          frame-locked image parity vs DuckStation
+tools/tests/test_gpu_frame.py
+```
+
+RetComM Studio's **Frames** tab is a viewer and launcher over exactly these
+artifacts. It never decodes a GP0 packet itself, so a headless capture and the
+GUI can never disagree about what a frame contained.
+
+---
+
+## Why this exists
+
+`gpu_frame_dump` on the debug server has always stamped every GP0 packet with
+the guest code that issued it — `func`, `pc`, `ra`, plus the linked-list OT rank
+(`handle_gpu_frame_dump` in `runtime/src/debug_server.c`, `GpuGp0RingEntry` in
+`runtime/include/gpu.h`). Nothing consumed that attribution, so "the glow is
+opaque and the vignette is missing" stayed a description of a screenshot rather
+than a pointer at a function.
+
+---
+
+## The provenance rule
+
+Everything these tools report is **observed** from one execution of one frame,
+and is labelled as such. It is never merged into a static claim.
+
+A `func` in a frame dump proves that code ran and issued a primitive. It says
+nothing about the call graph — the analyser's static coverage gap
+(`FUNCTION_DISCOVERY.md`) stays exactly as wide as it was. Studio shows observed
+counts and static names in different columns for the same reason.
+
+---
+
+## What is decoded faithfully, and what is not
+
+**Faithful:**
+
+* Vertices, including the 11-bit signed coordinate and the running GP0(E5) draw
+  offset — the same offset `gp0_exec_*` applies in `gpu.c`.
+* Packet word layouts for every polygon, line, rectangle, fill and copy form.
+* The texpage latch. A textured polygon's own tpage word overwrites draw-mode
+  state (`set_tpage_from_poly` in `gpu.c`), and later untextured polygons and
+  rectangles — which carry no tpage word — inherit it. Getting this wrong
+  mislabels precisely the semi-transparency mode you are usually chasing.
+* All four semi-transparency blends, when layers are rendered:
+  `0.5B+0.5F`, `B+F`, `B-F`, `B+0.25F` (GPUSTAT bits 5-6).
+
+**Not:**
+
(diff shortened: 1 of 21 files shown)


TechnicallyComputers:
PR summary

Title: cdrom: a Read that continues the current stream must not restart the drive

What was broken

Legend of Mana's land-creation effect rendered as hard-edged coloured wedges instead of a soft glow. The fault was in the CD-ROM controller, not the renderer.

The game streams a file by issuing Setloc(next)+ReadN about 10,800 guest cycles after each sector arrives. Every ReadN called start_read_stream(), which clears the sector buffer — destroying a sector that had already been delivered and announced but not yet drained by the guest.

┌──────────┬─────────────┬──────────────────────┐
│  buffer  │ psx-runtime │     DuckStation      │
├──────────┼─────────────┼──────────────────────┤
│ 0x0E1F18 │ 125111      │ 125111               │
├──────────┼─────────────┼──────────────────────┤
│ 0x0E2718 │ 125113      │ 125112 ← the palette │
├──────────┼─────────────┼──────────────────────┤
│ 0x0DCF18 │ 125113      │ 125113               │
└──────────┴─────────────┴──────────────────────┘

Sector 125112 carries the effect's 5-colour palette (0x0888F8 = (248,136,8), 0xB0F8F8 = (248,248,176), verified against the disc image). Destroyed before the guest could take it, its buffer received 125113's raw file bytes instead — which the colour routine then faithfully scaled into 216 distinct vertex colours. Those were the wedges.

The fix

DuckStation only calls BeginReading() — which clears buffers and re-arms the read-start latency — when the request actually moves the drive; otherwise it logs "Ignoring read command with pending/same setloc, already reading". A Read issued while the drive already streams the sector the pending Setloc names means "keep going", not "start over". The command is still ACKed with the same INT3; only the restart is skipped.

Two consequences: the un-drained sector survives, so the game receives the consecutive sectors it expects (109,110,111,112,113 rather than 109,111,111,113,113); and each request stops re-paying initial_read_delay_cycles(), so inter-sector gaps drop from ~456,000 cycles to 225,792 — exactly the 2× cadence. The load takes 2 guest frames instead of 13, matching DuckStation. int1_pended goes 10 → 0.

Also included

Developed alongside and verified in the same runs, all hardware/DuckStation-faithful:

- 8-slot sector buffer ring, as the hardware has. A delivered sector fills the next slot and the INT1 announcing it latches the read slot, so a sector arriving mid-drain no longer clobbers the FIFO. The read pointer follows the slot each INT1 announces, captured when the INT1 is raised — this drive free-runs, so latching "newest" at presentation time handed the guest a sector seven ahead of the one it asked for.
- The seamless no-IRQ refill advances the read pointer, since it continues a drain already in progress. Without it a multi-sector DMA starves mid-transfer (measured: 1,128,960 starved drains).
- Pended data-ready released on a short schedule rather than inside the guest's ack store, matching QueueDeliverAsyncInterrupt.
- Flow control for accelerated (non-1×) reads — never overwrite a sector the guest hasn't consumed while running faster than hardware, capped at the authentic sector period. Inert at authentic timing.

Savestate compatibility

The wire format is unchanged. The ring is transient device state, so only the slot being drained is serialised, in the pre-ring layout. Existing savestates load. Read-ahead slots and a pending Setloc are dropped across a load, costing a re-read the game already knows how to request.

This matters more than it appears: boot_state applies sections as it goes with no rollback, so a section failing its size check leaves RAM/CPU/GPU already overwritten — a format change here doesn't merely reject old states, it half-restores the machine. Worth hardening separately; this change stays inside the format to avoid tripping it.

How it was diagnosed

Everything above the CD was excluded by measurement first: renderer, display list, GTE, MIPS translation (lockstep clean over 5.6M blocks), fade scale, colour arithmetic, and the source colour table. Direct comparison against a patched DuckStation oracle then showed byte-identical CD request streams, with every request delivered exactly (delivered == requested, data matching the ISO, lost=0).

That left the transfers. Adding a matching CD DMA log to the oracle showed the two slot maps diverging at exactly one buffer; per-sector guest-cycle stamps showed the guest was never the slow party (it issued its next request within ~10,800 cycles every time) while ~451,584 cycles of read-start latency were being re-paid per request. Following that to start_read_stream()'s buffer clear gave the mechanism.

Regression tripwires

New cdrom_state counters ring_starved and ring_dropped, plus cdrom_timing_dump for per-sector deadline/delivery records. ring_starved and int1_lost must stay 0.

Verification

Legend of Mana (SLUS-01013): the effect renders correctly, the palette table holds 5 colour words (was 216), sectors arrive at the 2× cadence, and no notification is lost.

⚠️ This changes CD buffering for all games. The counters above are the first thing to check if another title regresses. Note also that this is a runtime change on a recompiler/config branch — if the stack ordering matters for merge, it may want relocating.

mstan:
Closing as superseded by narrower splits and already-landed runtime work. The CD-ROM behavior/diagnostics from this stack are already on master via ffa11fa2 (cdrom: a Read that continues the current stream must not restart the drive, validated for Legend of Mana). Remaining non-runtime pieces are now split as: #208 DuckStation oracle + GPU frame tools, #209 toolchain lazy urllib import, #210 setup-host runtime-dir staging, and draft #211 high-bank overlay/vault tooling. Original #169 remains open.

PR #189, opened 2026-08-23
presentation: reconstruct and filter FMV presents

mstan:
Split from original PR #169 by tetrisgm; original PR intentionally left open.

Scope:
- Sync the CPU VRAM mirror when entering 24-bit/15-bit display paths.
- Reconstruct low-res 24-bit FMV presents in the GL renderer before window scaling.
- Add the shared [video] fmv_filter setting and runtime/launcher round-trip conversion.

Kept out of this split:
- PAL timing and guest-clock changes.
- Turbo/load-acceleration/audio-sink changes.
- Texture-pack and PGXP changes.

Validation:
- python runtime/tests/test_fmv_quiet_guards.py
- cmake --build build\\split-runtime --target psx-runtime
- cmake --build build\\split-recompiler --target psxrecomp-toml

Notes:
- Added one split-only fix to bound launcher conversion against PSXRecompV4::VIDEO_FMV_FILTER_COUNT instead of an unreconciled launcher-only constant.

PR #204, opened 2026-08-25
gl: sync CPU mirror on 24-bit FMV entry

mstan:
Split out of the original WipEout 3 monolith (#169) and the presentation/FMV PR (#189).

This PR keeps only the 24-bit FMV CPU-mirror coherency fix:

- sync the CPU VRAM mirror once when entering 24-bit display mode
- avoid stale letterbox pixels when MDEC only updates the movie rows
- leave FMV filtering/reconstruction, diagnostics, and default presentation behavior out of this branch

Rationale:

WipEout 3's intro uses a 320x192 movie inside a 320x240 band. The movie rows are written by MDEC, but the bars can be left untouched. If the GPU FBO has pre-movie clears that have not been read back to the CPU mirror, the 24-bit present path can sample stale CPU mirror contents in those untouched rows.

Validation performed:

- `git diff --check`
- Tomba runtime configure/build against this psxrecomp branch with `PSX_REWIND=OFF` and `PSXRECOMP_ALLOW_NO_BIOS=ON`

Runtime visual validation still wanted before merge:

- WipEout 3 intro / letterbox case this was split for
- Tomba 1 FMV spot check
- Tomba 2 FMV spot check
- Mega Man X6 FMV/warning-adjacent spot check

### DIFF
--- a/runtime/src/gpu_gl_renderer.c
+++ b/runtime/src/gpu_gl_renderer.c
@@ -1409,6 +1409,10 @@ static void flush_pack_if_sampling(int tpage_x, int tpage_y, int depth,
 }
 
 /* ---- coherency: GPU -> CPU readback -------------------------------------- */
+/* Defined with the depth24 policy below; needed here for the readback guard. */
+static int s_depth24_skip_up = 0;
+static DirtyRect s_d24_skip_fb; /* union of skipped MDEC FB rects (VRAM halfwords) */
+
 static void ensure_cpu(void) {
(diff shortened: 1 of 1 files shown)


mstan:
Validation for this split from #169/#189: WipEout 3 Special Edition intro/FMVs spotchecked clean, including the target black letterbox areas with no stale pixels/flicker; Tomba 2 intro/FMVs spotchecked clean as a regression check. Proceeding with merge.

PR #216, opened 2026-08-25
config: honor title window width

tetrisgm:
## Summary

Allow a title's `game.toml` `[video]` block to declare its initial window width.

`settings.toml` already supports `window_width`, but the same key in the shipped title profile was silently ignored. This meant a title could declare its renderer, aspect, and supersampling while its intended initial window geometry was lost.

A positive title-profile width is treated as an explicit window request, so the existing fit-to-display maximize path does not replace it. Player `settings.toml` is loaded afterward and retains precedence. When the key is absent, the historical fit-to-display behavior is unchanged.

## Scope

- one optional `RuntimeConfig` field, defaulting to zero/inert
- `[video].window_width` parsing and range validation (640..7680)
- runtime application before player settings
- focused config regression coverage
- no height/UI schema expansion and no title-specific behavior
- no dependency on another unmerged PR

## Verification

- `video_enhancement_settings_test` passes, including valid and rejected-width cases
- fresh recompiler configure: 98 test files, all registered
- `cmake --build build-game-window-width --target video_enhancement_settings_test -j 4`
- `ctest --test-dir build-game-window-width -R video_enhancement_settings_test --output-on-failure`
- complete `psx-runtime` build succeeds from a fresh runtime configure

Verified on MinGW/GCC 16.1.0.

### DIFF
--- a/recompiler/src/config_loader.cpp
+++ b/recompiler/src/config_loader.cpp
@@ -545,6 +545,14 @@ static RuntimeConfig parse_runtime_block(const toml::value& cfg, const fs::path&
             }
             rt.video_supersampling = static_cast<int>(n);
         }
+        if (video.contains("window_width")) {
+            const auto n = toml::find<int64_t>(video, "window_width");
+            if (n < 640 || n > 7680) {
+                throw std::runtime_error(fmt::format(
+                    "[video] window_width out of range (640..7680): {}", n));
+            }
+            rt.video_window_width = static_cast<int>(n);
+        }
         if (video.contains("antialiasing")) {
             rt.video_antialiasing = toml::find<bool>(video, "antialiasing");
         }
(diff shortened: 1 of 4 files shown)


mstan:
Split validation note for #216:

- Rebased/cherry-picked cleanly onto current master after #225.
- `git diff --check origin/master...HEAD` passes.
- `video_enhancement_settings_test` passes, including `[video] window_width` parse/validation coverage.
- Checked suite title configs: Ape Escape, Tomba, Tomba 2, and MMX6 do not currently set `window_width`, so this PR has no shipped behavior impact for those titles unless a title explicitly opts in.
- Built Tomba against current master + #216, patched only the staged build copy of `game.toml` with `window_width = 1280`, and launched it. Manual validation confirmed it did not open as the 21:9/maximized window.

Scope note: this does not solve the separate default-window-size complaint. It only gives title configs an explicit width knob that opts out of the current maximize fallback.

PR #225, opened 2026-08-25
Treat RAM below a high-loading boot EXE as overlay region

TechnicallyComputers:
## Summary

A boot EXE that loads **high** and streams its gameplay code into the RAM
*beneath* itself had every one of those overlay targets refused by the
interpreter, falling through to `psx_unknown_dispatch` and a fail-fast
`exit(1)`.

The overlay-region model assumed the boot EXE sits at the bottom of RAM:
`[KERNEL_END, text_end)` is main-EXE text, `[text_end, RAM)` is runtime-loaded
overlay code. That holds for a `load_address` of `0x80010000`, where the text
base equals the kernel-window end and the region below the text is empty. It is
false for a high-loading title.

Klonoa loads a 46 KB stub at `0x80180000` (`text_size = 0xB000`) and runs its
overlays from `0x10000`–`0x130000`, so the floor landed at `0x18B000` and the
**entire** overlay region fell below it.

## Root cause

Two gates keyed off `g_overlay_region_floor` alone, and the floor is one-sided:

1. `dirty_ram_interp.c` — the admission heuristic for executable RAM that is not
   marked dirty requires `phys >= g_overlay_region_floor`. Overlay pages arrive
   by **CD DMA**, which never passes through the CPU write hooks, so they are
   never marked dirty. A `JALR` into one was refused and became a fatal
   unknown-dispatch.
2. `memory.c` — `dirty_ram_clear_image_baseline()` wiped the dirty bits across
   `[0x10000, 0x18B000)` believing that span was the clean compiled text image.

Traced on Klonoa (SLUS-00585). The failing call is an ordinary vtable dispatch:

```
0x80018018: lw   $v1, 2404($gp)
0x80018020: lw   $a3, 12($v1)          ; callback field -> 0x80123D00
0x80018028: beq  $a3, $zero, ...       ; null-guarded, so the game expects it live
0x8001805C: jalr $ra, $a3
```

and `0x80123D00` is real, valid code — an overlay relocation routine that reads a
descriptor and publishes section pointers. It is present in the region's
`executed_pcs` from a working run, so the code executes correctly whenever it is
allowed to run at all.

## The fix

Pin the text **base** (`g_text_image_lo`) alongside the floor and route both
gates through one predicate — the overlay region is both sides of the text image,
not just above it:

```c
static inline int phys_is_overlay_region(uint32_t phys) {
    return phys >= g_overlay_region_floor ||
           (phys >= DIRTY_RAM_KERNEL_WINDOW_END && phys < g_text_image_lo);
}
```

**No-op for bottom-loading games:** `g_text_image_lo` stays at its `0x10000`
default, the below-text clause is empty, and the baseline clear starts at the
same page as before.

Other titles that load high and should benefit: Street Fighter Alpha 3
(`0x80113B00`) and Bomberman Fantasy Race (`0x8003004C`).

## Verification

Klonoa, Linux, Release, OpenGL:

| | before | after |
|---|---|---|
| exit reason | `FAIL-FAST unknown dispatch @0x80123D00` | `atexit` (clean) |
| frame reached | 687–701, every run | 3364, still running when killed |
| dispatch misses | fatal on the first | **0 recorded for the whole run** (`unknown_dispatch_tail: total 0`) |
| crash file / freeze dump | both written | neither |

Startup now reports the pinned base:

```
psxrecomp: overlay_region_floor = 0x18B000 (game text end), text_image_lo = 0x180000
```

Independent of #226 (the Linux self-compile PR) — the two touch disjoint
files and can merge in either order.

### DIFF
--- a/runtime/include/dirty_ram_interp.h
+++ b/runtime/include/dirty_ram_interp.h
@@ -99,6 +99,34 @@ void dirty_ram_irq_ambient_resync_after_restore(void);
 extern uint32_t g_overlay_region_floor;
 #define OVERLAY_REGION_FLOOR (g_overlay_region_floor)
 
+/* The text-image BASE (phys) = this game's main-EXE load address. The floor
+ * above only bounds the text from ABOVE; on its own it encodes the assumption
+ * that the boot EXE sits at the bottom of usable RAM, so that [KERNEL_END,
+ * FLOOR) is text and [FLOOR, RAM) is overlay. That holds for a load_address of
+ * 0x80010000 (text base == KERNEL_END, so the region below text is empty) but
+ * is FALSE for a boot EXE that loads HIGH and streams its gameplay code into
+ * the RAM BELOW itself: Klonoa loads at 0x80180000 (text 0x180000-0x18B000,
+ * overlays at 0x10000-0x130000), Street Fighter Alpha 3 at 0x80113B00,
+ * Bomberman Fantasy Race at 0x8003004C. For those the sub-text RAM was
+ * misclassified as main-EXE text — clear_image_baseline() wiped its dirty bits
+ * and the dispatch admit-heuristic refused it, so a JALR into an overlay page
+ * the CD DMA'd there fell through to psx_unknown_dispatch and fail-fast
+ * exit(1) (Klonoa, frame 687, target 0x80123D00).
+ *
+ * The overlay region is therefore BOTH sides of the text image, not just above
+ * it. main.cpp pins g_text_image_lo = load_address & 0x1FFFFFFF at game load;
+ * the default equals DIRTY_RAM_KERNEL_WINDOW_END so the below-text clause is
+ * empty for BIOS-only runs and for every bottom-loading game. */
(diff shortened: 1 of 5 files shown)


mstan:
Split validation note for #225:

- Runtime focused tests pass: overlay capture/retry, overlay store cycle barrier, cause IP2 combinational, PSX cycle event boundaries.
- Source guard tests pass after adding coverage for the new high-load overlay-region predicate.
- Built no-regression smoke binaries successfully for Ape Escape, Tomba, Tomba 2, and Mega Man X6 against current master plus this PR.
- `test_interpreter_perf_guards.py` still fails on master and this branch with the existing `psx_devices_mmio_sync` inline-cycle-limit guard failure; not introduced by this PR.

Risk note: this is intended to be a no-op for bottom-loading games because `g_text_image_lo` stays at the kernel-window default, making the below-text overlay range empty. Positive validation still wants a high-loading title such as Klonoa, where the original failure was JALR into code loaded below the boot EXE text image.

PR #226, opened 2026-08-25
Self-compile hotfixes: offline dependency vendoring, GLVND GL, Windows/MinGW C++17, and a green ctest

### LINKED ISSUE #224: Building on Linux
### Summary

After a clean build (`recompiler/build-cli`, Linux), `ctest --output-on-failure` reports 3 failures. Two of them (`launcher_vulkan_option`, `vk_present_wait_stage`) look like the tests weren't updated after recent source refactors — the functionality they're checking for is present in the source, just under a different exact string than the test asserts on. The third (`aot_overlay_discovery`) looks like a separate, unrelated timing/locking issue in the test harness itself.

Filing all three together since they came from the same `ctest` run; happy to split into separate issues if preferred.

### Environment

- OS: Linux
- Build: `recompiler/build-cli`, built via `cmake --build .` / ninja
- `ctest --output-on-failure` result: 51 tests run, 13 initially failed due to unbuilt test executables (resolved by building all targets before running ctest — worth calling out in the README/CONTRIBUTING if it isn't already, since `ctest` doesn't build anything itself), leaving 3 genuine failures below.

---

### 1. `launcher_vulkan_option` — assertion string doesn't match current source

**Test:** `runtime/tests/test_launcher_vulkan_option.py`, line 24:
```python
assert "gi.renderer_labels      = kPsxRendererLabels;" in main
```

**Current source:** `runtime/src/main.cpp`, line 10321:
```cpp
gi->renderer_labels = kPsxRendererLabels;
```

Two mismatches:
- `gi` is now a pointer (`->`), not a value (`.`)
- The multi-space alignment in the test string is gone from the current formatting

This lines up with commit `9228f32c` — *"runtime: keep Vulkan opt-in and remove old launcher"* (2026-07-21) — which appears to have changed `gi` to a pointer and reformatted the surrounding code without updating this test's assertion string.

### 2. `vk_present_wait_stage` — assertion may be checking the wrong extracted region

**Test:** `runtime/tests/test_vk_present_wait_stage.py`, line 12:
```python
assert "VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT" in body
```

The constant **is** present in `runtime/src/gpu_vk_renderer.c` (lines 684, 1646, 1949, 1950), so a plain whole-file substring check should pass. Since it doesn't, `body` is likely extracted from a specific function or region (by name/regex) rather than the full file — and that region probably no longer matches post-refactor, likely from commit `7f9f59a2` — *"vulkan: wait for acquire before transfer writes"* (2026-07-14) — which touched this area.

I haven't dumped the exact extraction logic in the test yet; happy to attach the full `body`-slicing code from the test file if useful for triage.

### 3. `aot_overlay_discovery` — looks like a timing/locking assumption issue, possibly platform-specific

**Test:** `recompiler/tests/test_aot_overlay_discovery.py`

Failure output:
```
WARNING: ownership discovery cycle; suppressing 3 unstable derived roots (80010080,80010100,80010140)
Traceback (most recent call last):
  File "recompiler/tests/test_aot_overlay_discovery.py", line 3720, in <module>
    sys.exit(main())
  File "recompiler/tests/test_aot_overlay_discovery.py", line 3665, in main
    check_atomic_dll_publication()
  File "recompiler/tests/test_aot_overlay_discovery.py", line 1428, in check_atomic_dll_publication
    assert elapsed >= 0.6, f'{reader} escaped pair lock in {elapsed:.3f}s'
AssertionError: coverage escaped pair lock in 0.000s
Traceback (most recent call last):
  File "<string>", line 15, in <module>
    m.publish_shard_pair(sys.argv[2], sys.argv[3], sys.argv[4])
  File "tools/compile_overlays.py", line 4999, in publish_shard_pair
    os.replace(staged_dll, dll)
  File "<string>", line 9, in pause_after_manifest
    result = real(src, dst); count += 1
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpk_bb_e5s/.new.dll' -> '/tmp/tmpk_bb_e5s/00010000_DEADBEEF.dll'
```

The test expects a reader to be blocked out of a "pair lock" for ≥0.6s while a writer publishes a shard atomically via `os.replace`. Instead the reader gets through instantly (0.000s), and the writer itself then fails because the staged file it expects (`.new.dll`) is already gone by the time it tries to publish it — consistent with two processes racing on the same temp path with the intended lock not actually being held/respected on this platform.

Given the test monkeypatches `os.replace` (`pause_after_manifest`) to simulate delay, this may be a Linux-specific difference in how the lock/rename semantics are being emulated versus what the test assumes (possibly written/verified primarily on Windows). Haven't dug further into whether this is a real bug in `compile_overlays.py`'s locking or purely a test-harness assumption — flagging for someone closer to that code to assess.

---

### Suggested next step

For #1 and #2, if the underlying functionality is confirmed correct (renderer labels wired via pointer, present-wait stage present in the right function), these are just stale assertions and a one-line fix each. Happy to open a PR updating the two assertion strings if that's welcome — let me know.

For #3, more context on the intended locking behavior would help before proposing a fix.

TechnicallyComputers:
## Summary

Everything needed to clone this repo and build it yourself — on Linux/SteamOS
without a working network path to github.com, and on Windows/MinGW at all —
plus the three `ctest` failures reported in #224.

No engine behaviour changes here. The interpreter fix for high-loading boot EXEs
is split out into #225; the two touch **disjoint files** and can merge in either
order.

## 1. Offline dependency vendoring (`hotfix for offline build deps`)

`Build step for psx_libchdr failed: 1` behind a proxy or TLS-inspecting firewall
was the first thing a fresh checkout hit: libchdr is the first dependency the
runtime resolves and the only one with no local fallback (no `find_package`
path, no toolchain-pack copy), so a build with no network dies there.

- `third_party/deps.manifest` is now **the single record** of every source
  archive the build may fetch — name, vendored filename, sha256, upstream URL.
  Both `cmake/psx_dependency_archive.cmake` and `tools/ci/vendor_deps.sh` read
  it, so a vendored archive can never drift from the URL/hash the build would
  otherwise download.
- The pinned libchdr archive is committed under `third_party/`, so an
  up-to-date checkout does not download it at all.
- `tools/ci/vendor_deps.sh` stages every pinned archive, for teams that want a
  fully air-gapped tree (SDL3 and zlib resolve from a system package or a
  `cmake-clang-v1` toolchain pack first, so they are not committed).
- New `PSX_DEPS_OFFLINE` (or env `PSX_DEPS_OFFLINE=1`): never download. A
  dependency that is neither vendored nor pre-extracted becomes a **hard error
  naming the missing archive**, instead of a fetch that stalls behind a proxy.
- New `PSX_THIRD_PARTY_DIR` to relocate the vendored archive directory.
- `docs/BUILDING.md` documents both options and the failure mode;
  `THIRD_PARTY_ATTRIBUTION.md` covers the newly committed archive.

## 2. GLVND / Steam Deck GL resolution (`cmake patch hardening for steamos`)

`find_package(OpenGL)` reports OpenGL as **found** on a GLVND host that has
`libOpenGL.so` but no legacy `libGL.so` / `glx.h` — a Steam Deck — while never
creating the `OpenGL::GL` target. Hardcoding that target then fails at *generate*
time, before a single file compiles.

Resolution now goes through `recomp_resolve_gl()` (`recomp-ui/cmake/recomp_gl.cmake`),
which picks whatever the host actually has. With `PSX_RECOMP_UI=OFF` there is no
recomp-ui checkout to include, so the same resolution order is inlined:
`OpenGL::GL` → `OpenGL::OpenGL` → `OPENGL_gl_LIBRARY` → `OPENGL_opengl_LIBRARY`.

## 3. Windows/MinGW self-build

Two defects that only bite a MinGW C++17 build. Both are still live on `master`
today; the fix has been sitting on `feat/rbengine-bak` (`ae9432a3`) and never
reached `master`.

- `runtime.cmake` declared only `c_std_11` on the runtime target, leaving the
  **C++** dialect to whatever the parent project happened to set. A game
  CMakeLists that omits `CMAKE_CXX_STANDARD` therefore compiled
  `mod_packages.cpp` as a pre-17 dialect. Now states the runtime's real
  requirement (`cxx_std_17`) rather than inheriting it.
- `mod_packages.cpp` captured a **structured binding** inside a lambda. That is
  a C++20 extension: clang accepts it silently, MinGW/GCC at C++17 rejects it.
  Bound the map entry to ordinary locals before the lambda.

Cherry-picked from `ae9432a3`, minus that commit's unrelated `lib/recomp-net`
submodule bump. Verified by building `psx-runtime`.

## 4. Three `ctest` failures on Linux — fixes #224

All three are **test-only** defects. No product code is at fault. Details and
verification are in the commit message; briefly:

| test | actual cause |
|---|---|
| `launcher_vulkan_option` | `gi` became a pointer and the code was reformatted. **Two** assertions are stale, not one — `assert` stops at the first, and line 25 also renamed `vulkan_offered` → `vulkan_offered_b`. Blame is `61279a9d`, not `9228f32c` (which *added* the source and the assertions as a matching pair). |
| `vk_present_wait_stage` | Not a stale assertion. `652be7ba` added a forward declaration of `submit_present`, and the regex latches onto the **prototype**, so `body` became the *next* function (`finish_present`) and `submit_present` was never examined. `VK_PIPELINE_STAGE_TRANSFER_BIT` passed coincidentally on `finish_present`'s barrier — a false pass. Anchored the match on the definition. |
| `aot_overlay_discovery` | The test hardcodes `.dll`, but `load_region_coverage` re-derives the shard path via `overlay_ext()` = `.so` on Linux. Writer held `<stem>.dll.pair-lock`, reader took `<stem>.so.pair-lock` — different files, no contention, reader crossed the publication window in 0.000s. This is exactly why only *one* of the two readers failed: `shard_pair_files_complete` is handed the `.dll` path directly and locked correctly. |

The pair-lock machinery in `compile_overlays.py` is **sound** — `fcntl.flock`
works, `load_shard_func_ids` does take the lock, and all five renames occur in
the expected order with `#4 = .new.ranges → .ranges`. Only the test's naming was
wrong.

Verified on Linux: tests 1 and 2 pass; `check_atomic_dll_publication` now blocks
both readers (1.004 s / 1.016 s) and passes in 2.4 s.

### Known remaining, deliberately not fixed here

`aot_overlay_discovery` still fails **later**, in
`check_candidate_capacity_publication`, from the *same* class of defect —
**80** hardcoded `.dll` literals remain in that file and
`cache_candidate_inventory` also filters on `overlay_ext()`. That is a wider
Linux port of the test than #224 describes and wants an owner rather than a
blind sweep.

Related: `recompiler/CMakeLists.txt:485` states this test is `DISABLED`, but no
`set_tests_properties` marks it so — only the three known-failing guards are
actually disabled. Either the property or the comment should change; a red
registered test is precisely what `docs/TESTING.md` says destroys trust in the
suite.

## 5. Project scaffolding: ignore `analysis/`

`analysis/` is the documented default `--out` for `tools/gpu_frame_capture.py`,
`gpu_parity.py`, and `gpu_frame_layers.py`, and in practice also collects disc
asset rips. It routinely exceeds GitHub's 100 MB blob limit — a real title repo
here accumulated a 407 MB `STR00.BIN` and could not push at all — and rips are
disc content that must never be committed.

Added `/analysis/` to the new-project wizard template
(`templates/gitignore.in`, used by both `setup_project.sh` and `.ps1`), the CI
template (`docs/ci/templates/game.gitignore`), the migration tool's required-rule
list (`project_studio/ops.py`), and the audit check (`project_studio/detect.py`)
so existing repos are flagged and fixed by `project_studio apply`.

Verified end-to-end against a real title repo: `audit` reported
`[WARN] Missing patterns: /analysis/` → `apply --only merge_gitignore` →
`[OK] Essential setup-host ignore rules present.`

## 6. Submodule plumbing (the two `bump` commits)

`.gitmodules` pointed `lib/retcomm-rbengine` at `branch = master`, which no
longer exists — it is `main`. Plus routine pointer bumps for `lib/recomp-net`
and `lib/retcomm-rbengine`.

These are the least related commits in this PR; happy to drop them into a
separate one if you would rather keep the submodule pins under their own review.


mstan:
Validation note for #226 on current master after #216/#225:

- Actual no-commit merge result is clean and much smaller than the stale branch diff: dependency vendoring/CMake, three source-guard test fixes, `mod_packages.cpp` C++17 cleanup, `/analysis/` gitignore scaffolding, and docs. No analyzer/runtime feature rollback in the merge result.
- `git diff --cached --check` passes on the actual merge result.
- `tools/ci/vendor_deps.sh --check psx_libchdr` passes; committed libchdr archive SHA256 matches `third_party/deps.manifest`.
- Fixed one small `vendor_deps.sh --help` range bug so help no longer prints implementation code; pushed as c50f3f32.
- Runtime configured with `PSX_DEPS_OFFLINE=1`, `PSX_RECOMP_UI=OFF`, `PSX_REWIND=OFF`, `PSXRECOMP_ALLOW_NO_BIOS=ON`; configure explicitly selected the vendored libchdr archive and generated successfully.
- Built `psx-runtime` successfully from that offline/runtime config, exercising the CMake dependency helper, inline OpenGL resolution path, and the `mod_packages.cpp` C++17 compile fix.
- Recompiler configured and `psxrecomp-game` built successfully.
- Direct changed guard tests passed: `test_vk_present_wait_stage.py`, `test_launcher_vulkan_option.py` under `PYTHONUTF8=1` on Windows, and focused `check_atomic_dll_publication()` from `test_aot_overlay_discovery.py`.
- `tools/new_project_layout/project_studio` Python files compile with `py -3 -m compileall`.

Caveat: full Linux `test_aot_overlay_discovery.py` still has the wider known `.dll`/`.so` cleanup called out in the PR body; this PR only fixes the reported atomic-publication check.

PR #227, opened 2026-08-25
Self-build: pack sysroot into CMake and require toolchain 1.0.14

TechnicallyComputers:
## Summary

Three self-build fixes recovered from `feat/rbengine-bak`, where they have been
sitting unmerged. Each is real, each is currently broken on `master`, and none
of them is engine behaviour.

Separate from #226 on purpose — that one is already under review, and these
touch **disjoint files** (`psxrecomp_cli.py`, `host/psxrecomp_codegen_host.c`,
`tools/toolchain_pack.py`, `recompiler/CMakeLists.txt`; #226 touches
`runtime/runtime.cmake`, `runtime/chd_dependency.cmake`, `cmake/`,
`third_party/`, tests and project templates). The two can merge in either order.

## 1. Wire the pack sysroot into CMake — SteamOS self-build

Cherry-picked from `0bb45442`.

`clang.cfg` already hands the *compiler* a `--sysroot`, but CMake's `find_*`
still searches the host. On a headerless host — stock SteamOS — recomp-ui's
`find_package(OpenGL REQUIRED)` therefore dies at configure time:

```
Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR)
```

`_pack_sysroot_cmake_args()` re-roots `find_*` into the pack with
`CMAKE_SYSROOT`. `OpenGL_GL_PREFERENCE=LEGACY` is required alongside it, not
optional: the sysroot ships legacy `libGL.so` + `GL/gl.h` but no GLVND dev files
(`glx.h` / `libOpenGL.so` / `libGLX.so`), so the GLVND path cannot complete
inside the sysroot and `OpenGL::GL` would never be created. It no-ops off Linux
and when no pack sysroot is present.

> **Reviewer note — interacts with #226.** #226 fixes a *different* GL failure
> (a GLVND host where `find_package(OpenGL)` reports found but never creates the
> target). This one forces `LEGACY` preference. Both are real and they sit at
> different layers, but I have **not** verified the two compose on a host that
> hits both. Worth a check before the second of the two lands.

The commit's `extra_cmake_args` line was dropped in the conflict resolution:
that parameter is `rbengine`-only and does not exist in this function on
`master`.

## 2. Toolchain min-version floor 1.0.14

Cherry-picked from `00f0a9a9`.

v1.0.14 is the first `cmake-clang-v1` pack carrying the Linux build sysroot.
Older cached packs compile against host headers that do not exist on stock
SteamOS, and their objects **mis-link after a pack upgrade**
(`__isoc23_strtoul`). A cached pack is otherwise accepted forever once any cmake
runs, so a version floor is the only mechanism that ever replaces a stale one.

`RETCOMM_TOOLCHAIN_MIN_VERSION` still overrides in either direction;
`0` / `off` / `none` disables the floor entirely (pre-floor behaviour).

Lands in two files that deliberately mirror each other —
`tools/toolchain_pack.py` and the C copy in `host/psxrecomp_codegen_host.c` used
by the runtime's own self-compile path. They must move together or they drift.

## 3. `PSXRECOMP_ENABLE_CHD` (default ON)

Hand-extracted from `ea1b157d`.

libchdr is needed by exactly one target here — the user-facing `psxrecomp` media
CLI — and arrives via FetchContent + ExternalProject, the heaviest and most
fragile step in the configure. Two concrete failures:

- it needs the network on a cold cache; and
- **ExternalProject cannot handle a source path containing a colon.** A repo
  directory such as `Marvel vs. Capcom: Clash of Super Heroes Recomp` truncates
  to `.../Marvel vs. Capcom.rule` and configure dies. Nothing platform-specific
  about it — it depends on the title's name.

Gating the include and the target on one option lets a caller who only wants the
emitters configure without any of that. Default ON, so existing invocations are
unchanged.

Hand-extracted rather than cherry-picked because `ea1b157d` also introduces the
`psxrecomp-analyze` target and its `analysis_db` / `analysis_export` sources,
which do not exist on `master`.

Verified both ways on Linux:

| | result |
|---|---|
| `-DPSXRECOMP_ENABLE_CHD=OFF` | configures clean; `psxrecomp-game` builds and links; media CLI correctly absent (`ninja: error: unknown target 'psxrecomp'`) |
| default (ON) | `PSXRECOMP_ENABLE_CHD:BOOL=ON` in the cache; media CLI target present, builds and links |

## Deliberately left behind

`feat/rbengine-bak` carries 291 commits not on `master`. I swept it for
build/portability fixes and promoted only what is genuinely build-scoped and
extractable. Not brought over, because each needs sources that do not exist on
`master` or is feature work that merely touches a build file:

- SIO1 serial-link test targets (`runtime/CMakeLists.txt`) — needs `src/sio1.c`,
  `src/psx_link.c`
- `psxrecomp-analyze` target + CLI analyzer plumbing — function-discovery feature
- `PGXP_CLONE` / `_pgxp` executable suffix — PGXP A/B variant feature
- framework-owned mod catalog staging (8 MB RAM mod) — feature

Also checked and **not** a gap: the `mods/packages` clear-scope fix and
`PSX_SDL3_SOURCE_DIR` are already on `master`.

One pointing the other way: `recomp_resolve_gl` (#226's GLVND fix) is **not** on
`feat/rbengine-bak`, so a careless future merge of that branch would regress it.

### DIFF
--- a/host/psxrecomp_codegen_host.c
+++ b/host/psxrecomp_codegen_host.c
@@ -1496,6 +1496,55 @@ static void handle_progress_line(const char* line,
     }
 }
 
+/* Last-lines capture so a failed CLI run can say WHY in the wizard, which has
+ * no console: the child's stderr shares the progress pipe, JSON progress rows
+ * contribute their "message", raw rows (compiler/traceback text) contribute
+ * as-is, and the most recent error-looking line is appended to err_msg. This
+ * is what turns "psxrecomp rebuild failed (exit 1)" into "… failed (exit 1):
+ * Could NOT find OpenGL (missing: OPENGL_INCLUDE_DIR)". */
+typedef struct {
+    char last[480];
+    char last_err[480];
+} CliTail;
+
+static int cli_tail_line_is_error(const char* s) {
+    return strstr(s, "rror") != NULL || strstr(s, "ailed") != NULL ||
+           strstr(s, "FAILED") != NULL || strstr(s, "Traceback") != NULL ||
+           strstr(s, "fatal") != NULL || strstr(s, "Fatal") != NULL;
+}
+
+static void cli_tail_note(CliTail* t, const char* line) {
+    char msg[480];
+    const char* rec = line;
+    int is_error_event = 0;
+    if (line[0] == '{') {
+        /* sdk_progress emits compact JSON: {"event":"error","message":…}. */
+        is_error_event = strstr(line, "\"event\":\"error\"") != NULL;
+        if (!json_get_string(line, "message", msg, sizeof(msg)))
+            return; /* structured row without text (e.g. result) */
+        rec = msg;
+    }
+    if (!rec[0])
+        return;
+    snprintf(t->last, sizeof(t->last), "%s", rec);
+    if (is_error_event || cli_tail_line_is_error(rec))
+        snprintf(t->last_err, sizeof(t->last_err), "%s", rec);
+}
+
+static void cli_fail_msg(char* err_msg, size_t err_cap, const char* fail_label,
+                         long code, const CliTail* t) {
+    const char* why = t->last_err[0] ? t->last_err : t->last;
+    if (code == 3) {
+        snprintf(err_msg, err_cap, "Disc verification failed (wrong dump).");
+        return;
+    }
+    if (why[0])
+        snprintf(err_msg, err_cap, "%s failed (exit %ld): %s", fail_label,
+                 code, why);
+    else
+        snprintf(err_msg, err_cap, "%s failed (exit %ld).", fail_label, code);
+}
+
 #if defined(_WIN32)
 static int run_cli_win(const char* cmdline,
                        RecompLauncherCPrepareProgressFn on_progress,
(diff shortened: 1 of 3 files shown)


mstan:
Rebased onto current master after #210/#212/#228 landed. The optional-libchdr CMake option is already present on master now, so this PR's active diff is reduced to the self-build/sysroot/toolchain-floor pieces plus CLI failure reporting. Also added a small portability fix so the host C code uses _stricmp on Windows instead of calling strcasecmp directly.\n\nValidation on the rebased branch:\n- git diff --check origin/master...HEAD\n- py -3 -m py_compile psxrecomp_cli.py tools/toolchain_pack.py\n- toolchain_pack default floor smoke: default 1.0.14; 0/off/none disable; explicit override accepted\n- recompiler configure with -DPSXRECOMP_ENABLE_CHD=OFF\n- CHD-off build: psxrecomp-game and psxrecomp-analyze\n- CHD-off target list confirms psxrecomp media CLI absent\n- default configure has PSXRECOMP_ENABLE_CHD=ON\n- default build: psxrecomp\n- CTest: cli_project_packaging

PR #233, opened 2026-08-25
input: stop folding the D-pad onto the analog stick

TechnicallyComputers:
## Summary

In pinned-ANALOG mode the runtime presented a single D-pad press as **two
inputs at once**: `pad_sticks_for` slammed the left stick to full deflection
while `pad_buttons_for` went on reporting the D-pad bits. A real DualShock
holds both sticks at `0x80` while the D-pad is down, so this is a state no
hardware produces — and an analog-aware game that reads both sources acts on
one press twice.

Reported as: single D-pad taps registering as double inputs in Legend of Mana.

## Root cause

`runtime/src/main.cpp`, `pad_sticks_for`, reached because Player 1 defaults to
`PAD_MODE_ANALOG`:

```c
if (dpad_pressed("left"))  out[0] = 0x00;
if (dpad_pressed("right")) out[0] = 0xFF;
if (dpad_pressed("up"))    out[1] = 0x00;
if (dpad_pressed("down"))  out[1] = 0xFF;
```

The fold was deliberate — it let stick-only movement paths respond to the
D-pad — and its comment described it as "the faithful 'seamless' behaviour".
The sibling fold on the injection path went further and claimed it was "same
as a real DualShock seat". That claim is false, and it is what licensed the
controller-side fold.

## Evidence

Measured on Legend of Mana's land-placement screen, narrowing from the host
pad down to the write that moved the cursor:

| Probe | Finding |
|---|---|
| `pad_trace` | Host presents **one** press per tap — no bounce, no analog flip, steady 60 fps — but carrying **two representations**, D-pad bit plus full stick deflection, arriving in the same frame (`skew=0`) |
| `pad_delivery` | SIO delivers one contiguous 6-poll run per tap, polls exactly 10 bytes apart; `up`/`down` bit-for-bit mirrors (`0xFFEF`+`ly=0x00` vs `0xFFBF`+`ly=0xFF`). Delivery is clean |
| `menu_scan` | Cursor index at `0x800DCEEC` responded to 14 of 15 presses, stepping **two** entries on 6 of them. The double-buffered display copies at `0x8004C035`/`0x8004C14D` (0x118 apart) mirror it at 3x scale |
| `cursor_writers` | Doubled presses ran the **same handler twice** — `pc=0x8006583C` (decrement) / `0x800658C0` (increment), overlay `func=0x00000F40`, identical return address — on consecutive 30 Hz logic ticks. Frame gaps always even (2 or 4), never odd |

Everything upstream of the fold measured clean and is not implicated.

**Controlled test.** DIGITAL mode changes two variables at once (stick bytes
*and* pad ID `0x73`→`0x41`), so it only showed "not analog-with-fold" fixes it.
This change alters **only the stick bytes**, leaving the pad ID at `0x73`:

- Before, ANALOG: cursor stepped two entries on 6 of 14 presses
- After, ANALOG (`pad_status` → `analog=True`, `sticks=[128,128,128,128]`):
  **33 presses, 33 single steps, zero doubles**

That exonerates the pad type and confirms the fold specifically. Confirmed
visually by the reporter under deliberate stress testing.

## Compatibility

**Games whose movement reads only stick magnitude (e.g. Tomba) will no longer
respond to the D-pad in pinned-ANALOG.** The removed code named Tomba
explicitly. Such games are served by **HYBRID** mode, where pressing the D-pad
flips the pad to digital so the game runs its own d-pad path — without ever
presenting the impossible both-at-once state. This is a mode change for
affected users, not a silent equivalence, and `TombaRecomp` is worth a check
before this lands.

The keyboard path is unaffected: `psx_keybinds_sticks` maps that player's bound
stick-direction keys and is their only stick source.

## Scope

The injected-input fold in `apply_input_override_to_sio` is **retained**. It is
unreachable by a player, and injection has no physical stick to steer with, so
removing it would break tooling that drives stick-only games via `set_input`.
Its comment no longer claims to match hardware. Happy to remove it too if
preferred — two lines at each of two sites.

## Testing

- `ninja -C build-release Legend_of_Mana__Recompiled` — clean
- 177 tool unit tests passing
- Live capture above, in ANALOG mode, on the affected screen

## Also included

Four diagnostic probes with unit tests (`776ebc07`), each answering a question
the previous could not: `pad_trace` (what the runtime presents), `pad_delivery`
(what the game reads, via the always-on SIO ring), `menu_scan` (which RAM
location is the cursor), `cursor_writers` (who wrote it, grouped per press —
separating "two handlers ran" from "one handler ran twice" from "one write of
two steps"). They can be dropped from this PR if you'd rather keep the hotfix
minimal.

### DIFF
--- a/runtime/src/main.cpp
+++ b/runtime/src/main.cpp
@@ -4147,18 +4147,23 @@ static uint16_t pad_buttons_for(const PlayerInput& p, int player, bool suppress_
 
 /* Analog stick bytes (lx,ly,rx,ry) for a player; centred if no live source.
  *
- * The host left stick maps to the LEFT analog axes (variable). When fold_dpad
- * is set, the physical D-pad (and, for a keyboard player, the arrow keys) is
- * ALSO folded onto the left axes at full deflection, so an analog-mode game
- * whose movement reads only the stick magnitude (e.g. Tomba) still responds to
- * the D-pad/keyboard — the faithful "seamless" behaviour: the stick gives
- * variable speed, the D-pad gives full speed, both at once. Only the PHYSICAL
- * D-pad is folded in (not the stick->d-pad button mapping), so the stick keeps
- * its true variable magnitude. fold_dpad is false in HYBRID mode (there the
- * D-pad instead flips the pad to digital, so the game runs its own d-pad path)
- * and true in pinned-ANALOG mode. The keyboard branch always folds the arrows
- * — they are that player's only stick source. */
-static void pad_sticks_for(const PlayerInput& p, int player, uint8_t out[4], bool fold_dpad) {
+ * The host left stick maps to the LEFT analog axes (variable). The physical
+ * D-pad is deliberately NOT folded onto those axes: a real DualShock holds
+ * both sticks at 0x80 while the D-pad is pressed, so presenting a direction
+ * as a button bit AND a full stick deflection at once is a state no hardware
+ * produces. An analog-aware game that reads both sources then acts on one
+ * press twice — measured on Legend of Mana's land-placement cursor, which
+ * stepped two entries per tap in pinned-ANALOG and exactly one in DIGITAL.
+ *
+ * A game whose movement reads only the stick magnitude (e.g. Tomba) is served
+ * by HYBRID mode, where pressing the D-pad flips the pad to digital and the
+ * game runs its own d-pad path — without ever presenting the impossible
+ * both-at-once state.
+ *
+ * The keyboard branch is unaffected: psx_keybinds_sticks maps that player's
+ * bound stick-direction keys onto the axes, which is their only stick
+ * source. */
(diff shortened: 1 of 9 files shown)


mstan:
Closed as superseded by #235.

What was accepted from this work:
- physical D-pad is no longer folded onto the left analog stick in pinned analog mode, keeping the presented pad state closer to real DualShock behavior and avoiding the Legend of Mana double-input issue described here
- the pad/cursor diagnostic tools and tests were preserved

What changed before merge:
- the global/public Hybrid controller mode was not accepted as framework architecture
- #235 replaces it with a trusted game-plugin controller presentation callback, so titles like Tomba can keep their Hybrid Controller mod behavior without making hybrid mode a global launcher/game.toml surface

Validated before merge:
- psxrecomp controller guard tests and pad/cursor tool tests passed
- Tomba was regenerated and rebuilt against rbengine main
- Tomba Hybrid Controller behavior was manually validated with the slot-2 100% save

PR #234, opened 2026-08-26
tools: fix macOS CI packaging failure — empty-array expansion under set -u

TechnicallyComputers:
## Summary

Fixes the **macOS CI workflow failure** in the packaging step:

```
psxrecomp/tools/package_setup_host.sh: line 268:
  RUNTIME_DIRS[@]: unbound variable
Error: Process completed with exit code 1
```

(One accuracy note: the failure is in the setup-host packager shell script
invoked by the workflow, not in CMake itself — the CMake configure/build
completes before this point.)

## Root cause

macOS runners still ship **bash 3.2** as `/bin/bash`. Under `set -u`, bash 3.2
treats `"${arr[@]}"` on an **empty** array as an unbound variable and aborts.
bash >= 4.4 does not. The script runs `set -euo pipefail`, so this class of bug
is completely invisible on Linux and Windows and only ever breaks macOS CI.

`RUNTIME_DIRS` is populated solely from `--runtime-dir`, so **any game packaged
without one aborts**. Introduced in `05cd3625` ("tools: stage runtime dirs in
setup-host packages", 2026-08-24).

## Audit

Rather than patch the one reported line, every shell tool under `set -u` was
audited: 15 array expansions, of which **7** can legitimately be empty at the
point of expansion.

| Script | Array | Status |
|---|---|---|
| `package_setup_host.sh` | `RUNTIME_DIRS` | **the CI failure** |
| `package_setup_host.sh` | `RUNTIME_DIRS_OPTIONAL`, `PROJECT_FILES`, `PROJECT_DIRS` | same shape — the check at line 112 only rejects *both* project arrays being empty, so either alone can be |
| `package_release_macos.sh` | `REQUIRE_MODS` | latent, in a **macOS-only** script |
| `bundle_mingw_dlls.sh` | `REQUIRE` | latent |
| `bios_emitter_fingerprint.sh` | `present` | latent — empty when no listed file exists |

Left alone as already safe: five sites already using the guard; both
`vendor_deps.sh` loops, which are guarded by a `${#WANTED[@]}` check one to two
lines above; and `args` / `stage_args`, which are unconditionally populated
before use.

## Fix

All seven now use the `${arr[@]+"${arr[@]}"}` idiom **already used elsewhere in
these same scripts** (`bundle_mingw_dlls.sh`, `stage_setup_sdk.sh`), so this is
consistent with the existing style rather than a new convention. A comment at
the packager records why the quoting looks odd, so it does not get "cleaned up"
later.

## Regression guard

Adds `tools/tests/test_shell_nounset.py`, which scans every `set -u` shell
script for this pattern and fails on a new one — the rule lives in the artifact
instead of in review memory. A `${#arr[@]}` size check within three lines counts
as a guard, which is how `vendor_deps.sh` protects its loops. It includes a test
that the detector itself can fire, since a guard that cannot fail is not a
guard.

## Testing

- `bash -n` on all four modified scripts
- Full tool suite: 98 tests passing
- Idiom verified: empty and undeclared both expand to zero iterations; elements
  containing **spaces** survive intact (this repo's own checkout path has
  spaces); glob characters are not expanded; and guarded output is byte-identical
  to unguarded when the array is populated

**Limitation, stated plainly:** the bash 3.2 abort is **not reproducible on the
dev machine** — bash 5.3 tolerates both empty and undeclared arrays under `-u`.
The first version of the verification asserted that the unguarded form aborts,
and it did not; that test was wrong, not the code. So this fix rests on the
mechanism matching the CI error text exactly and on the idiom being provably
behaviour-preserving — **only the macOS runner can confirm it end to end.**

### DIFF
--- a/tools/bios_emitter_fingerprint.sh
+++ b/tools/bios_emitter_fingerprint.sh
@@ -63,6 +63,6 @@ for f in "${FILES[@]}"; do [ -f "$f" ] && present+=("$f"); done
 # records with a relative path while runtime.cmake recomputes with an absolute
 # one, and every configure read as STALE with a byte-identical tree. The array
 # order is defined by this script alone, so it is stable across spellings.
-for f in "${present[@]}"; do
+for f in ${present[@]+"${present[@]}"}; do
     sha256sum < "$f"
 done | sha256sum | awk '{print $1}'
--- a/tools/bundle_mingw_dlls.sh
+++ b/tools/bundle_mingw_dlls.sh
@@ -327,7 +327,7 @@ for i in "${!EXE_PATHS[@]}"; do
   bundle_one "${EXE_PATHS[$i]}" "${DEST_DIRS[$i]}" "${LABELS[$i]}"
 done
 
-for dll in "${REQUIRE[@]}"; do
+for dll in ${REQUIRE[@]+"${REQUIRE[@]}"}; do
   [[ -n "${dll}" ]] || continue
   for i in "${!EXE_PATHS[@]}"; do
     exe="${EXE_PATHS[$i]}"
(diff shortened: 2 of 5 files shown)


TechnicallyComputers:
yup 236 solves it.

PR #242, opened 2026-08-26
runtime: overlay_region_floor game.toml key (follow-up to #160)

Alexbeav:
## Summary

#160 added the PSX_OVERLAY_REGION_FLOOR environment override for titles whose secondary EXEs load below the boot text end. This exposes the same floor as a [runtime] overlay_region_floor key in game.toml so a title can declare it without an environment variable (Gran Turismo: GTOS/GTMENU/GTMAIN load at 0x80010000). The commit also lets a bundled overlay toolchain use gcc when it is on PATH; that half can be split out if preferred.

## Commits

- d596b0ec runtime: [runtime] overlay_region_floor config key; bundled toolchain drives gcc when available

## Scope

 6 files changed, 107 insertions(+), 31 deletions(-)

## Validation

Cherry-picked from the Alexbeav/psxrecomp `main` line, where the same changes pass the recompiler/runtime/runtime-ui ctest gates with only the pre-existing failures (`aot_overlay_discovery`, `release_zip`, `gte_register_access_test` link). This branch was also built on `mstan/master` as of 01c647e8 with the same result.

### DIFF
--- a/docs/config_schema.md
+++ b/docs/config_schema.md
@@ -330,6 +330,11 @@ The other load-time accelerators are likewise opt-in:
 [runtime]
 idle_skip = true
 turbo_audio_sink = true
+overlay_region_floor = "0x10000"   # optional: lowest RAM address treated as overlay region.
+                                    # Default = boot EXE text end. Lower it for titles whose gameplay
+                                    # code loads at/inside the boot text range (GT1 secondary EXEs at
+                                    # 0x80010000, Driver 2 mission pages) so it is overlay-cache
+                                    # eligible. Clamped >= 0x10000; PSX_OVERLAY_REGION_FLOOR overrides.
 ```
 
 ### `turbo_loads` / `offer_turbo_loads` — deprecated and ignored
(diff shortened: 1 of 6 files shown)


mstan:
Small config robustness concern.

`overlay_region_floor` accepts integer TOML values, but the parser casts `v.as_integer()` directly to `uint32_t` and leaves validation to the later runtime clamp/mask. That means negative or very large TOML integers can silently wrap/mask into a different physical address instead of failing loudly.

I think we should reject integer values outside the intended PSX physical/kernel-safe range at parse time, matching the fail-loud style used by nearby config fields. Do you agree? The string hex path already goes through `parse_hex`; this is only about the raw integer branch.

Alexbeav:
Agreed. Pushed 699022e7: raw TOML integers are now validated at parse time — negative or >32-bit values are rejected, and the KSEG-stripped value must name main RAM above the kernel window (0x10000..0x1FFFFF physical), failing loud in the same style as `debug_port`. The hex-string path still goes through `parse_hex` and now shares the same range check. `test_overlay_region_floor_config.py` pins both rejections.

PR #243, opened 2026-08-26
bios: replace the stale SCPH5552 profile

Alexbeav:
## Summary

`bios/SCPH5552.toml` on `master` predates the address-model schema. This replaces it with a profile in the current schema. It does not add support for another BIOS.

The `master` file describes the image with `bios_vectors` and `bios_aliases`. It has no `address_model` copy windows, no `install_slots`, no `runtime_exports` and no `[program.image]` pin, although `bios/SCPH1001.toml` uses all four and the recompiler reads them. Without the pin, the declared-identity gate cannot refuse a wrong revision.

## What changes

One file. The new profile matches `SCPH1001.toml` key for key:

- `[program.image]` pins SHA-256 `1faaa18f…` (v3.0 01/06/97, CRC32 `D786F0B9`)
- `address_model` and `runtime_exports` are unchanged from SCPH1001
- the five `install_slots` carry over with `len` and `resume`
- `rom` keeps `master`'s region-qualified name; `bios_rom_alias.h` resolves either convention

## Evidence

A direct comparison of the two dumps, each matching its declared identity (SCPH-1001 CRC32 `37157331`, SHA-256 `71af94d1…`; SCPH-5552 CRC32 `D786F0B9`, SHA-256 `1faaa18f…`):

| ROM range | Result |
|---|---|
| `0x00000–0x0FFFF` boot + Kernel Part 1 | identical |
| `0x10000–0x17FFF` Kernel Part 2 | identical |
| `0x18000–0x42FFF` shell | differs in 155,874 bytes |

Each install slot maps through Kernel Part 2 into identical bytes. Every seed in `phase2_ghidra_seeds_SCPH5552.json` lies below ROM `0x18000`, so the kernel-only corpus still holds.

Checked statically: the TOML parses, its section and key structure equals `SCPH1001.toml`'s, and `config_loader.cpp` reads every key it uses.

**Not run:** a generate or build on `master` with this profile.

## Evidence limit

The install slots were measured on a live SCPH-1001 boot. They are carried here on the byte identity above, not re-measured on a SCPH-5552 boot. The profile states that beside them.

## Out of scope: supporting another retail BIOS

The earlier review objection stands and this does not address it. A further retail image needs per-BIOS savestate isolation, a generalised netplay handshake, and launcher and recomp-ui changes. None of that is here.

This PR only corrects a file that is already on `master`: the CLI profile table maps `SCPH5552` to it (`main_cli.cpp:158`) and the packaging test requires it. If `master` should not offer SCPH5552 at all, that is a separate decision, and the file should then be removed rather than left stale.

## Aside

A comment in `recompiler/include/bios_rom_alias.h`, and the matching fixture comment in `runtime/tests/test_bios_rom_alias.cpp`, says `master`'s SCPH5552.toml names its ROM without `bios/`. The current file includes `bios/`, so that comment was already stale. It is not changed here.

### DIFF
--- a/bios/SCPH5552.toml
+++ b/bios/SCPH5552.toml
@@ -1,0 +1,86 @@
+# bios/SCPH5552.toml — BIOS build profile for the retail Sony SCPH-5502/5552
+# (Europe) v3.0 01/06/97 image ("System ROM Version 3.0 01/06/97 E",
+# SHA-256 1faaa18fa820a0225e488d9f086296b8e6c46df739666093987ff7d8fd352c09).
+#
+# Derived from bios/SCPH1001.toml. ROM 0x0000-0x17FFF (boot + Kernel Part 1/2)
+# is byte-identical to the US SCPH-1001 v2.2 image, so every kernel anchor
+# below (copy windows, install slot, deliver_event_ret, shell entry) carries
+# over unchanged. The shell (ROM 0x18000+) DIFFERS from the US image; this
+# profile therefore has no shell seeds (recompiler/seeds/
+# phase2_ghidra_seeds_SCPH5552.json = the SCPH1001 corpus filtered through
+# tools/filter_bios_seeds.py, kernel region only) and the shell window runs
+# through the dirty-RAM interpreter (build with -DPSX_SHELLWIN_INTERP=ON) until
+# an EU-shell seed corpus exists.
+#
+# The .BIN image itself stays untracked (bios/* gitignore rule) — profiles are
+# the only tracked contents of this directory.
+
+[program]
+name         = "Sony SCPH-5552 BIOS (Europe, v3.0 01/06/97)"
+id           = "SCPH-5552"
+rom          = "bios/SCPH5552.BIN"
+load_address = "0xBFC00000"
+entry_pc     = "0xBFC00000"
+text_size    = "0x80000"
+
+[recompiler]
+seeds    = "recompiler/seeds/phase2_ghidra_seeds_SCPH5552.json"
+out_dir  = "generated"
+out_stem = "SCPH5552"
+strict   = true
+
+# The BIOS's boot-time bulk copies of CODE out of ROM into RAM. Single source
+# of truth for every relocation window in the recompiler AND in the C it
+# emits (the generated dispatch's normalize(), the kernel-bless guard, the
+# game/shell-overlap tests). Each entry is a CLAIM that the copy is
+# byte-verbatim; the runtime's kernel-bless memcmp enforces it for the
+# bless window. All bounds are [lo, hi) with hi EXCLUSIVE.
+#
+# dispatch_key = "ram": functions in the window are keyed by RAM address
+#                       (normalize folds ROM -> RAM).
+# dispatch_key = "rom": functions keep their ROM key (normalize folds the
+#                       RAM alias back to ROM).
+[recompiler.address_model]
+normalize_mask = "0x1FFFFFFF"
+
+[[recompiler.address_model.copy]]
+name         = "Kernel Part 2"
+rom_lo       = "0x1FC10000"
+rom_hi       = "0x1FC18000"
+ram_lo       = "0x00000500"
+runtime_base = "0x00000500"     # KUSEG — the kernel executes uncached-region code cached at 0x500
+dispatch_key = "ram"
+kernel_bless = true
+
+[[recompiler.address_model.copy]]
+name         = "Shell"
+rom_lo       = "0x1FC18000"
+rom_hi       = "0x1FC43000"
+ram_lo       = "0x00030000"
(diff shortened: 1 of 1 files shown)


Alexbeav:
Converting this to a draft after CobaltCryptid's point on Discord, which is correct: this PR only adds the profile toml and does not touch the default `PSXRECOMP_BIOS_STEMS`, but the runtime currently treats BIOS identity as a two-way choice — savestates are filed under `openbios/` when the checksum matches OpenBIOS and under `scph1001/` otherwise, and the lobby handshake only carries `can_scph1001` / `prefer_openbios`. A title built against a third BIOS would misfile its savestates and be invisible to the netplay fallback.

Supporting an additional BIOS properly means at least:

1. savestate folder / compatibility keyed by a BIOS token or checksum rather than the OpenBIOS-or-SCPH1001 test;
2. a generalized lobby handshake (advertise the BIOS set each peer can run, fall back to OpenBIOS when the sets do not intersect);
3. detection of which BIOS(es) an executable was compiled with, so the runtime can pick the matching state folder and handshake answer.

Two ways forward — happy to do either:

- park this PR until those layers are generalized (I keep the profile in my fork for local PAL builds only), or
- extend this PR to the three items above, if you would rather see it done in one piece.

Leaving it as a draft until you say which.

TechnicallyComputers:
I would not merge this yet, if we are going to add support for yet another BIOS bin, we need to implement new handshake behavior in netplay lobbies, new behavior in recomp-ui, new behavior in the launcher...  Decide if/how to handle or support multiple retail BIOS in one executable for hotswapping and fallback...

Gamestates and RAM do not match between BIOS versions, which breaks savestates, requiring isolation of savestates in separate folders per-bios, and it also breaks netplay determinism for the same reason, which we currently handle by forcing a handshake between all peers about what BIOS are compiled into their executable, with support for hotswapping between SCPH1001.bin, and openbios - using OpenBIOS as a safe fallback everyone has access to if some peers do not match the retail bios selected by the host.

All of that has to be refactored at retcomm-launcher, recomp-ui, and psxrecomp in order to satisfy the request to add support for a new BIOS.  I suggest leaving this open until we can circle back to it.

Alexbeav:
@TechnicallyComputers I tested this before replying. The head (`a1885e4f4`) changes one file, `bios/SCPH5552.toml`, and it doesn't add a BIOS: master already maps SCPH5552 in the CLI profile table (`main_cli.cpp:158`), packages the file, and lists its CRC in `psx_bios_known_images.h`. The old TOML just predates the address-model schema.

Setup: Tekken 3 built on this head with OpenBIOS, SCPH1001 and SCPH5552 linked, one Windows peer and one Rocky 9 peer, matches started from the launcher lobby. SCPH5552 generated from this profile and linked on both, so the "not run" line in the PR body is covered now.

| Host / guest BIOS | Lobby said | Peers booted | Result |
|---|---|---|---|
| SCPH1001 / SCPH1001 | SCPH1001 | 1001 / 1001 | synced |
| SCPH5552 / SCPH1001 | "All users agree on proprietary BIOS SCPH1001.bin" | 5552 / 1001 | `rb boot dig0 MISMATCH`, ~20 s black window, both kicked to the lobby |
| SCPH1001 / SCPH5552 | same | 1001 / 5552 | same |
| OpenBIOS / SCPH1001 | OpenBIOS | OpenBIOS / OpenBIOS | synced |

So you're right that the lobby isn't ready for a second retail image. The settle can't tell SCPH5552 from SCPH1001: `can_scph1001` is set when any non-bundled retail dump is found (`main.cpp:9444-9458`, and `discover_retail_bios_near` walks up from the exe), and the settled token is only `openbios` or `scph1001`. The rollback boot digest check stops the match, but the player gets a black window and a kick with no reason. Lobby matches always ran rollback here; I didn't test a mixed pair in delay mode.

Savestates, unchanged by this PR: a SCPH5552 build files slots under `scph1001/` (`main.cpp:15547`). Each `.pst` records the BIOS checksum and the loader refuses a mismatch (`boot_state.c:773`), so a wrong-BIOS state is rejected, not loaded.

I'd keep this PR to the profile and do the multi-BIOS work as a follow-up:

1. The lobby offer carries the identity of the BIOS file each peer would actually boot (the selected file's `image_id` / CRC), not a yes/no for "some retail dump exists".
2. The room shows each peer's BIOS and flags a mismatch before the host can start, with OpenBIOS as the fallback when the sets don't intersect.
3. If the boot digest check still trips, the runtime says why instead of dropping back to the lobby silently.
4. Savestate folders keyed by that identity instead of `openbios` / `scph1001`.

recomp-ui and the launcher would follow the same offer format. If you'd rather master not offer SCPH5552 until then, I can change this PR to remove it from the CLI table and packaging instead of leaving a stale profile. If the split works, I'll open an issue for the follow-up.

PR #259, opened 2026-08-29
Multi-disc: per-disc identity, launcher roster, and stop copying dumps into projects

TechnicallyComputers:
Host side of mstan/recomp-ui#40, plus two fixes that multi-disc exposed.

## Multi-disc roster + selection

`game.toml [game] discs` now reaches the launcher as a roster, so the player can pick which disc Play boots. The choice persists as `settings.toml [disc] selected` (1-based).

`selected` **names** the disc; `path` only survives when it *is* that disc (same file-name stem). That ordering is the point of storing the index: an external launcher — or a hand edit — changes discs by writing one integer and it takes effect even though `settings.toml` still carries the previous disc's path. A player who browsed for their own copy of the selected disc is still honoured, because a relocated or container-swapped image keeps its stem. Single-disc titles are returned unchanged.

## Per-disc identity gating

Each disc of a set carries its own serial and its own TOC. Both gates were checking the whole set against disc 1's, so selecting disc 2 reported the player's correct disc as the wrong game, and refused it for online play.

- **`[game] disc_serials`** — array parallel to `discs`. Resolved through one shared `expected_serial_for_disc()` used by the launcher verdict, the launch-time disc check, and the netplay mount gate, so they cannot drift.
- **`[netplay] required_disc_fps`** — array parallel to `discs`. Resolution returns the whole `NetplayDiscExpect` per image rather than patching one string: `required_tracks` and `required_leadout_lba` are per-disc facts too (a set can mix a CD-DA disc with a data-only one, and the lead-out LBA is the disc's size). Only the fingerprint has per-disc data today; the resolution point now exists for the others.

A disc with no entry in either array is not gated at all rather than gated against another disc's number — the ISO-header check still applies. Absent arrays (every single-disc title, every port predating this) keep exactly the flat behaviour.

## Setup no longer copies discs into the project

`setup_project.sh` has always defaulted to `STAGE_DISC=0`. Two callers overrode it: `NewProjectOptions.stage_disc` defaulted `True` and Project Studio's checkbox was pre-ticked. Staging copies every track into the repo, so a 3-disc PS1 set duplicates ~2.2 GB of a ROM library the user already owns — for no gain, since `disc/` is gitignored and a staged copy is never shipped. The checkbox is relabelled "Stage disc" → "Copy disc into project"; "stage" did not say what it costs. `--stage-disc` still works.

## Window icon path

Publishes the path `psx_apply_window_icon()` already resolves so recomp-ui can put the same art on the launcher window, instead of the launcher repeating the search and drifting from it.

## Verification

Driven end to end on a 3-disc set sourced from an external dump library (no in-repo copy): all three discs list and verify with their own serials, Play boots the selected one, and hand-editing `selected = 3` while `path` still pointed at disc 2 moved the launcher to disc 3. Netplay is not compiled into that build, so the fingerprint gate was verified by resolving `game.toml` through the built `config_loader` and checking each disc lands on its own fingerprint — including the `.bin` spelling of a `.cue` roster entry — not by running an online session.

## Setup wizard: ask for every disc, and record them where the roster is read

Everything above makes the *launcher* multi-disc aware. The first-run **setup
wizard** was not, and the two are different code paths: the roster is built in
the runtime, while the wizard runs through `psxrecomp_codegen_host.c`, which
never read `[game] discs` at all. So a player scaffolding a set was asked for a
single image and found out much later that discs 2 and 3 were never recorded.

The setup host now reads `[game] discs` and publishes `gi.discs` /
`gi.num_discs`, on the same rule the runtime uses — a roster only appears when
there is more than one image, so single-disc titles are untouched. The reader is
~60 lines rather than a TOML dependency, recognises only that one key, and
absolutises relative entries against the project root (the launcher
existence-checks what we publish and its cwd is not necessarily that root).

Implements `persist_setup_discs` (recomp-ui#40, appended additively to
`RecompLauncherCGameInfo`), so the flush carries every located image instead of
only the mounted one. `disc.cfg` becomes one path per line in disc order — a
reader taking only the first line still gets disc 1, which is what it got
before. An unlocated slot is written as an empty line, so line N stays disc N
and filling one in later cannot renumber the rest.

### The part that actually closes the gap

Recording picks in `disc.cfg` is not enough, and this is worth stating plainly
because it is easy to assume otherwise:

```
roster Play selects from  <-  game_discs = gc.discs  <-  game.toml [game] discs
disc.cfg                  <-  read_cached_path()     <-  getline: FIRST LINE ONLY
```

`disc.cfg` is the mounted-image cache. The roster comes from `game.toml`. So the
wizard's picks reach the roster only by being written there — which is exactly
what the Project Studio path already did by running `probe_disc.py` per image
and `verify_disc_set.py` over the results. That tooling was not in this repo:
`verify_disc_set.py` was absent, and `probe_disc.py` was a stale fork 90 lines
behind, predating multi-disc support. Both are now vendored from retcomm-studio
(MIT into PolyForm-NC; notice and pin in `THIRD_PARTY_ATTRIBUTION.md`). The sync
drops nothing — the fork's only 8 unique lines were older revisions of the same
functions, and a single-disc project still renders `disc = "..."` byte-identically.

`update_disc_set.py` is new here, and is the piece neither repo had.
`probe_disc.py --write-game-toml` renders a **complete** `game.toml`: right when
scaffolding, destructive afterwards, since it would discard a live project's
`[video]`, `[controller]`, `[widescreen]`, `[netplay]` and every comment. This
probes each located image, verifies them as one set, then edits **only**
`disc` / `discs` / `disc_serials` / `required_disc_fps`. It skips any key whose
value it already agrees with, so it will not reflow a hand-formatted array and a
re-run is a no-op. It refuses to write when `verify_disc_set.py` rejects the
set — images that are not one program are a mistake to surface, not a config to
record. Generate runs it before invoking the CLI, non-fatally: Generate needs
only the boot image, and refusing to build because disc 3 would not probe is the
worse outcome.

The probe dumps it feeds to `verify_disc_set.py` go to a temp directory and are
discarded (`--artifact-dir` keeps them). Writing them beside `game.toml`, as it
first did, modified three **tracked** files in the game repo on every Generate —
`disc_probe*.json` are committed provenance records of the original scaffold,
and the rewrite also downgraded a curated `display_name` back to the raw file
name.

### Fix: both tools reported a status that had gone stale

`probe_disc.py` and `verify_disc_set.py` both told the user that *"the runtime
mounts the boot disc only today"* and named MULTI_DISC.md **P1 + P3** as
outstanding. P1 is what the rest of this PR does, so that sentence had outlived
itself — and it was wrong in the damaging direction, telling someone scaffolding
a set that the config they had just written did nothing.

Corrected to what is true: `discs` is a first-class config entry, the runtime
builds a roster from it and mounts the **selected** disc, remembered in
`disc_index`. What is still outstanding is narrower than the old text implied —
specifically **P3**, a faithful lid in `cdrom.c`. Choosing a disc means choosing
it before the game starts; swapping one mid-session is not implemented. Worth
saying explicitly, because the roster this tooling writes is easy to mistake for
in-game hot-swap.

Fixed upstream in retcomm-studio first and re-synced rather than patched here,
so the vendored copies do not fork from the MIT original the attribution pins.

### Verification of this half

Driven against a real 3-disc set (Final Fantasy VII, SCUS-94163):

- The setup host publishes `num_discs = 3` with `persist_setup_discs` set and
  all three absolute paths in disc order, by driving
  `psxrecomp_codegen_host_apply` directly against that project's `game.toml`.
- From a **single-disc** `game.toml`, `update_disc_set.py` produces `discs`,
  `disc_serials` (`SCUS-94163/94164/94165`) and `required_disc_fps` identical to
  the values Project Studio had already produced for that project — and leaves
  all 114 remaining lines of config and comments byte-for-byte intact. A second
  run changes nothing, and `--check` agrees.
- The cherry-pick onto this branch was recompiled at this base with the project
  toolchain; all three Python tools parse.

Not verified: the wizard driving this end to end on screen. Each stage was
exercised in isolation, not as one first-run flow.

https://claude.ai/code/session_0194TYJAqgyG9a9PJe9uUcgQ

---

## Folded in: host frame pacing and the hotkey focus gate

`feat/host-pacing-and-hotkey-focus` is merged into this branch, which supersedes
**#260** (`fix/host-frame-pacing`). Two runtime commits:

- **don't fast-forward while the window lacks input focus** — a held turbo/
  fast-forward hotkey kept applying after focus moved away, so the guest ran
  unbounded in the background.
- **pace the frame even when driver vsync should own the cadence** — when the
  driver did not actually throttle (vsync off, or a compositor ignoring it),
  nothing else bounded the frame, so guest speed ran away.

They land here rather than separately because both branches already shared this
branch's base (`d14ca691`, `c481b6cd`, `10f443eb`); the merge adds only those
two commits and touches only `runtime/src/main.cpp` (+52 lines), with no
conflicts against the multi-disc work.

#260's two commits are **patch-identical** to these — same `git patch-id` on
both pairs, only the SHAs differ from being rebased onto master directly — so
closing #260 loses nothing. Verified that this branch now contains
`feat/multi-disc-launcher` and `feat/host-pacing-and-hotkey-focus` as ancestors,
and every #260 patch by patch-id.

---

## Also folded in: packaging gates (#267) and the setup-host branch

`fix/package-src-mods` (**#267**) and `feat/setup-host-multi-disc` are merged
here. Both touch `tools/package_setup_host.sh` and the packaging template, so
landing them separately would have meant three PRs racing on the same files.

**From #267 — two pre-zip gates and a Studio audit.** A release zip could ship
a `CMakeLists.txt` naming a file the zip did not contain; the failure only
surfaced on a user's machine. Final Fantasy VII hit it twice in a row:
`src/ff7_mods.c` missing (fails at configure), then `../psx_symbols.h` missing
(fails at the first object file).

- The template stages `src/`, `mods/` and the generated symbol headers, each
  guarded on existence — `copy_proj()` hard-fails on a missing path.
- `package_setup_host.sh` refuses to zip a stage whose `CMakeLists.txt` names
  an unstaged path, or whose **project sources include one**. The second gate
  matters because a relative `#include` is a compile input nothing in
  `CMakeLists.txt` names. Guarded `if(EXISTS ...)` paths, globs, cmake-variable
  paths, and headers reachable through `-I` out of the staged framework trees
  are all correctly not flagged.
- Project Studio's Migrate tab gains `packager_stages_sources` (FAIL) and
  `packager_stages_optional` (WARN), plus `op_sync_packager_project_dirs`,
  which patches an existing wrapper rather than regenerating it — packagers get
  hand-extended with extra `--project-file` lines a rewrite would drop.

This is a class problem, not a one-off: of 23 game repos with a packaging
script, all 23 carry `mods/` and 3 stage it; 7 carry `src/` and 2 stage it. The
three that do are hand-patched copies of the same fix.

**From `feat/setup-host-multi-disc`** — only two patches were not already here
(verified by `git patch-id`): a revised "ship the mod catalog in every release"
and a setup-host fix. Everything else was the same content under different SHAs.

**One regression the merge introduced, and fixed in `29cba084`.** Both branches
added `src/` to the template — the setup-host branch unconditionally in the
static argument list, #267 as a guarded `EXTRA_PROJECT` entry. git merged them
cleanly because they touch different lines, leaving `src/` passed twice. The
duplicate is harmless; the unconditional entry is not, because `copy_proj()`
hard-fails on a missing path, so any project without `src/` would stop
packaging with `error: missing src`. The static entry is dropped and the
guarded one kept.

Verified after both merges: all three merged branches are ancestors, both gates
are present in `package_setup_host.sh`, the template still stages every base
entry, the shell and Python compile, and the Studio audit still detects and
fixes the real Final Fantasy VII gap.


TechnicallyComputers:
Added a packaging commit to this branch: **ship the mod catalog in every release, and let Studio bundle one**.

Separate concern from multi-disc, but it lands here because the disc/setup-host work in this PR is what surfaced it.

### The bug

A built title stages its mod catalog next to the exe, but nothing put it in the shipped package — every release went out with an empty Mods page. The packager already had `--runtime-dir mods`; it was opt-in and `RUNTIME_DIRS` defaulted to empty, so each title had to pass it from its own generated wrapper. An audit found **21 of 23 ports never did**, and the two that did used the `-optional` variant that warns instead of failing.

### What changed

- **`package_setup_host.sh`** stages `mods/` by default and *requires* it — `runtime.cmake` guarantees `<exe>/mods` for every non-oracle target, so a missing one means broken build wiring and should fail the release. `--no-mods` opts out; an explicit `--runtime-dir-optional mods` still wins. It also ships the source catalog, because the setup-host zip rebuilds on the player's machine and that rebuild re-runs POST_BUILD staging from the source tree.
- **`package_release.ps1`** (Windows/MinGW) had no mod handling at all. Now stages before the stray-file scan so the existing gate covers it, strips the build machine's `state.toml`, asserts ≥1 manifest survived.
- **`package_release_macos.sh`** shipped **one** package where the player should have seen five: it took the catalog from `mods/preloaded` "never the build tree", but framework builtins (PGXP, fast loading, CD speed, bezel) live in `psxrecomp/mods/builtin` and only ever appear next to the exe. The real hazard was `state.toml` inside the build tree, not the build tree — so take it and delete that one file.
- **Wrapper template** gains `--project-dir src`, so titles ship game-owned trusted-plugin sources; `CMakeLists.txt` references them unconditionally and the rebuild can't configure without them.
- **Studio can now bundle.** There was no export action — the only route to a distributable was install-CI, commit, push, wait, which is how an empty Mods page survived unnoticed. `buildops.export_release()` runs the *same wrapper CI runs*, exposed as `build export` on the CLI and an "Export release zip" button in the GUI.

### Verification

Packaged real zips rather than reading code. Default produces 5 catalogs (`ff7.enhancements`, PGXP, fast-loading, cd-speed, bezel) with no `state.toml`; `--no-mods` reproduces the old behaviour at 0.

### Note for reviewers

Existing ports will now **fail packaging** if their build didn't stage `<exe>/mods`. That's the intended gate, but it's a behaviour change across ~21 repos — expect it on their next release rather than be surprised by it.

https://claude.ai/code/session_0194TYJAqgyG9a9PJe9uUcgQ

TechnicallyComputers:
Follow-up commit on this branch: **a developer channel, kept out of published releases**.

Mods under active development had nowhere to live — a package was either in the catalog and therefore in every release, or not built at all.

### How it works

A package declares its own maturity:

```toml
format_version = 5
id = "example.enhancements"
channel = "developer"
```

It still ships with ordinary local builds and local exports, so it can be tested exactly as a player would see it. Release packaging drops it.

**The exclusion defaults from `$CI`** — on under any CI provider, off locally. That's deliberate: a per-workflow opt-in is precisely the shape that left 21 of 23 ports shipping no mods at all, so this must not require 23 release workflows to be edited before it takes effect. `--exclude-dev-mods` / `--include-dev-mods` and `EXCLUDE_DEV_MODS=0|1` override either way, and `project_studio build export --exclude-dev-mods` reproduces what a release contains.

### Two details worth review

- **Both catalogs are pruned.** The setup-host zip ships the source catalog under `mods/preloaded` for the on-machine rebuild. Pruning only `mods/packages` would publish the package in source form and let the player's own rebuild put it straight back.
- **Manifests are never rewritten during packaging.** Editing a package while packaging it would change content that's meant to be verifiable, so the marker is package-level and pruning removes the whole package directory. Split a player-ready feature into its own stable package rather than promoting a catalog that still holds in-progress work.

Applied in all three packagers (setup-host, Windows, macOS); each asserts no developer manifest survived.

### Verification

| | catalogs | dev package present |
|---|---|---|
| local (no `$CI`) | 5 | yes |
| `CI=true` | 4 | no |
| `CI=true EXCLUDE_DEV_MODS=0` | 5 | yes |

The release gate was also tested both ways: it passes on a CI zip and correctly trips on a local one. A gate that can't fire is worth nothing.

https://claude.ai/code/session_0194TYJAqgyG9a9PJe9uUcgQ

PR #260, opened 2026-08-29
runtime: two causes of unbounded guest speed (present pacing, turbo focus gate)

TechnicallyComputers:
Two independent causes of the same reported symptom: the guest running far
above 1.0x with nothing held, indistinguishable to a player from a stuck
fast-forward. Both are host-side present/input bugs; neither touches codegen.

### 1. Hold-to-turbo ran while the window had no input focus

`SDL_GetKeyboardState()` returns a **cached** array that SDL reconciles from
events. A key whose key-UP is delivered to a different window stays latched
there with nothing to clear it — and the canonical way to move focus away is
Alt+TAB, where TAB is exactly the fast-forward bind. The guest then sits at the
fast-forward multiplier indefinitely. Regaining focus makes SDL reconcile the
array, which is why the symptom appears to stop the moment you click back in
and press anything. (The report that surfaced this blamed the FPS-readout key
for "stopping the turbo"; it was the focus change that came with reaching for
it.)

Gated the held-key read on real input focus. This is independently correct —
hold-to-turbo must not run while the player is typing in another application.
Edge-triggered hotkeys are unaffected: they go through `host_keymap_match()` on
real key **events**, which are only delivered to the focused window anyway.
Headless and pre-window cases are explicitly not suppressed.

### 2. The wall-clock pacer was skipped whenever vsync "owned" the cadence

`present_should_wall_pace()` returned early on `present_vsync_owns_cadence()`,
trusting the GL swap to block. **A swap that does not block then leaves the
guest with no speed limit at all.**

Running the pacer underneath vsync is safe because `frame_pacer_wait()` is
deadline-based rather than a fixed sleep: when the swap has already consumed
the period, `now >= next_deadline` and it returns immediately after advancing
the deadline. So it costs nothing where vsync works — in particular it adds no
latency to the healthy path, which is why vsync was given sole ownership in the
first place — and supplies the cap where vsync does not block.

The frame-interpolation exclusion is deliberately kept: that path subdivides
this same interval and would genuinely double-pace.

The startup log line is corrected with it. `wall-clock pacer skipped` is how a
bring-up session decides pacing is accounted for, and it is what sent this
investigation hunting a phantom stuck fast-forward instead of an unpaced
present.

### Measurements

FF7 (SCUS-94163), NVIDIA GL under a compositing WM, 45 s, nothing held,
telemetry off:

| state | guest frames | fps | speed |
|---|---|---|---|
| before either fix | 9678 | — | — |
| after the focus gate | 4771 | ~106 | 1.77x |
| `PSX_VSYNC=0` (pacer forced on) | 2651 | ~58.9 | 1.00x |
| **both fixes, default settings** | **2643** | **~58.7** | **0.98x** |

The two are sequential findings on one symptom: the second is only visible
once the first stops masking it, which is why the frame counts chain. Turbo
attribution was confirmed by changing `PSX_FAST_FORWARD_SPEED`, which only
`manual_fast_forward_multiplier()` reads and which is only reached inside the
gated branch — the speed tracked it, so the branch was demonstrably executing.

### Scope, verification, and the pin

- **Game-agnostic.** Runtime only, no per-game configuration, no new settings.
- **Pin bump required to consume; regen NOT required** — both changes are
  runtime-only (`runtime/src/main.cpp`), no codegen or emitter change.
- **Built and run in a real integration.** The identical changes are the
  submodule pin of
  [TechnicallyComputers/Final-Fantasy-VII@de2a9a2](https://github.com/TechnicallyComputers/Final-Fantasy-VII/commit/de2a9a2),
  which builds and boots (OpenGL, HLE BIOS, headless and windowed).
- This branch is those two commits **rebased onto current `master`**; the
  cherry-pick applied cleanly and the result compiles with the project
  toolchain (cmake-clang-v1). The frame measurements above were taken on the
  game's pin, not on this rebase.

### Known limitations

- Frame counts are from one machine and one GPU/WM combination. The failing
  case is specifically a non-blocking swap; on a driver where vsync genuinely
  blocks, change 2 is a no-op by construction rather than by measurement.
- Not oracle-diffed against Beetle. These are host present/input paths and do
  not alter guest-visible timing, interrupts, or emulated state, so a co-sim
  diff has nothing to compare — worth saying explicitly given CONTRIBUTING's
  verification bar.

### On PR focus

CONTRIBUTING asks for one thing per PR, and this is two fixes. They are here
together because they are one investigation of one symptom and the evidence
chains through both — the 4771-frame row is the state *between* them. Happy to
split into two stacked PRs if you would rather review them separately.

https://claude.ai/code/session_01UWiDNcWuNtDWHE7Pnhq4jp

### DIFF
--- a/runtime/src/main.cpp
+++ b/runtime/src/main.cpp
@@ -573,6 +573,30 @@ static void fps_telemetry_toggle(void) {
     host_osd_push(enabled ? "FPS readout on" : "FPS readout off", 1500);
 }
 
+/* Hold-to-act host hotkeys must not fire while the game window does not hold
+ * keyboard focus.
+ *
+ * SDL_GetKeyboardState() returns a CACHED array that SDL updates from events.
+ * A key whose key-UP is delivered to some other window stays latched in that
+ * array with nothing to clear it -- and the canonical way to move focus away
+ * is Alt+TAB, where TAB is exactly the fast-forward bind. The game then runs
+ * at the fast-forward multiplier with no key held, indefinitely. Regaining
+ * focus makes SDL reconcile the array, which is why the symptom "stops" the
+ * moment you click back into the window and press anything.
+ *
+ * Gating on real input focus fixes that and is independently correct: a
+ * hold-to-turbo must not run while the player is typing in another
+ * application. Edge-triggered hotkeys are unaffected -- they go through
+ * host_keymap_match() on real key EVENTS, which are only delivered to the
+ * focused window in the first place.
+ *
+ * No window (headless, or before the window exists) means focus is not a
+ * concept here; do not suppress in that case. */
+static bool host_hotkey_input_focused(void) {
+    if (!sdl_window) return true;
+    return (SDL_GetWindowFlags(sdl_window) & SDL_WINDOW_INPUT_FOCUS) != 0;
+}
+
 static int manual_fast_forward_multiplier(void) {
     static int value = -2; /* -2 = unread, -1 = max/unbounded */
     if (value == -2) {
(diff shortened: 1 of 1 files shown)


TechnicallyComputers:
Superseded by #259 — closing, not dropping.

Both commits here are now on `feat/multi-disc-launcher`, folded in by merging
`feat/host-pacing-and-hotkey-focus` (which carried the same two changes on top
of that branch's base rather than on master directly):

- runtime: don't fast-forward while the window lacks input focus
- runtime: pace the frame even when driver vsync should own the cadence

They are **patch-identical** to this PR's commits — `git patch-id --stable`
gives `4fd44880…` and `0e8fbc4c…` for both pairs; only the SHAs differ from the
rebase. Nothing is lost by closing this.

Reviewing them together is the point: the pacing work and the multi-disc host
changes both touch `runtime/src/main.cpp`, and #259 already contained the shared
base (`d14ca691`, `c481b6cd`, `10f443eb`), so keeping them apart meant two PRs
that could not merge independently anyway.

PR #269, opened 2026-08-30
runtime: gate save-state menu on rewind backing

mstan:
## Summary
- refuse opening the save-state menu when local rewind/save-state backing is unavailable
- block non-netplay save/load submissions with an actionable OSD before misleading slot/action handling
- keep the existing netplay host save-state path outside this local rewind gate

## Validation
- git diff --check
- py -3 runtime\\tests\\test_savestate_status_protocol.py
- cmake --build F:\\Projects\\psxrecomp\\_build-pr259-40-title\\mmx6 --target psx-runtime --parallel 6

Tracked in beads-eio.3.87.


mstan:
Closing this PR because the rewind-gating premise was wrong: one-shot disk save states should remain available even when rewind/rbengine is disabled. The actual Tomba 2 save-state fix was landed directly on master as 47bda817 (runtime: fix savestates at dirty boundaries).

PR #273, opened 2026-08-30
runtime: pace frames the GL renderer elides a swap for

kerokline:
The GL renderer may skip SDL_GL_SwapWindow when the displayed rect is unchanged, so a double-buffered 30 Hz game does not swap an identical front buffer on the intervening vblank. The frontend calls present once per guest vblank, so eliding a swap also elides whatever that swap was blocking on.

That is only safe while the wall-clock pacer owns cadence, because the pacer runs per present regardless of what the renderer does. When driver vsync owns cadence instead (~60 Hz panel, present_vsync_owns_cadence()), the pacer is deliberately not run -- the two are XOR to avoid double-blocking -- and the swap block becomes the only throttle. An elided frame then blocks on nothing, and a game presenting every other vblank advances two guest vblanks per block: exactly 2x real time.

Measured on BreathOfFire3Recomp (SLPS-00990, 60 Hz panel, OpenGL): guest vblank counter sampled against wall clock ran 120.4 fps / 2.01x before, 59.8 fps / 0.997x after. Swaps per second are unchanged at 60 either way -- the fix removes a free-running guest frame, not a swap -- and per-frame GPU cost is unchanged (present_gpu_ms_avg 12.2 -> 12.6, scene_gpu_ms_avg 4.5 -> 4.1).

Gate both skip sites (VRAM and wide-FBO paths) on the cadence owner, alongside the existing OSD and interpolation-cadence guards. Only the GL renderer has this optimisation; the software and Vulkan paths always present and were never affected.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -483,6 +483,9 @@ if(BUILD_TESTING)
         add_test(NAME frame_interpolation_context_ownership_test
                  COMMAND ${Python3_EXECUTABLE}
                          ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_frame_interpolation_context_ownership.py)
+        add_test(NAME present_skip_cadence_guard_test
+                 COMMAND ${Python3_EXECUTABLE}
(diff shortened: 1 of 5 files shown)


mstan:
Thank you for the contribution!

PR #282, opened 2026-08-30
runtime: correct CD-ROM seek and read handoff

Alexbeav:
## Summary

- model `Setloc` followed by `ReadN` or `ReadS` as an implicit seek
- keep `SEEK` active during the bounded delay, then deliver the requested target sector first and enter `READ`
- make explicit `SeekL` and `SeekP` cancel the old read stream, pending data-ready interrupt, sector ring, and `BFRD` request
- preserve the pending Setloc state through snapshots without changing the 4,931-byte CD-ROM snapshot size
- clear the pending state on controller reset
- add one compiled controller-level regression to both runtime and recompiler CMake suites

## Validation

- `cdrom_seek_semantics_test` passes in both CMake trees. It compiles and calls the production controller code.
- The test covers explicit-seek cleanup, implicit-seek delay, `SEEK` to `READ` status, target-first delivery, reset, snapshot round trip, old packed `0/1` compatibility, and the exact snapshot byte count.
- Clean recompiler and runtime product builds link on Windows with GCC 16.1.0 and Ninja.
- The external framework suite passes 54 of 56 enabled tests. The two failures are pre-existing Windows portability cases: one hard-coded MSYS2 GCC path and one LF-versus-CRLF ZIP expectation. Three tests are disabled.
- Fork review PR [Alexbeav/psxrecomp#15](https://github.com/Alexbeav/psxrecomp/pull/15) reviewed exact head `d43442bfe7f66ecc9f416c930321e2e10653bda1` twice and reported no issues.

## Owned-disc route

Duke Nukem: Land of the Babes (`SLUS-01002`) is the retail consumer. No retail data is included in this PR.

A fresh headless product built from this exact head has SHA-256 `6F388FD4D27CAFF67D89EE4A436C98BC81D12B031948A517A41B8BFEB18BB0CE`.

The accepted test state came from a newer portfolio save-state format (version 6), while this upstream base accepts version 5. For the upstream-base gate, I used a disposable copy with only the header version changed from 6 to 5. The section count remained 17; the version-5 reader explicitly skips the newer scheduler section. The original state was unchanged. This is bounded route evidence, not a general save-state migration claim.

Both exact input routes reached the complete first room with player control:

- slot 3, Cross for 10 polls, neutral for 2,000 polls, Start for 10 polls
- slot 3, Cross for 10 polls, neutral for 16,000 polls

The Start-skip route issued `Setloc 04:21:31`. Its first matching CD transfer used LBA 19,456. The first transfer word was `0x02312104`, which contains the same `04:21:31` sector header. The runtime reported no fatal state.


mstan:
We had to revert this via PR #287 because it regressed Ape Escape texture loading in the attract/demo path.

Bisect result:
- Ape known submodule pin `c5b8ccd` = good
- `ac9be0b6` / PR #277 = good
- first bad commit: `3b837f4e2d96a6e60f103cfd1b5cf5059cb6b349` (`Fix CD-ROM seek and read handoff`)
- current master plus the PR #282 revert was rebuilt and visually confirmed fixed for Ape Escape

A follow-up PR is welcome, but it needs to preserve Ape Escape behavior. Please include an Ape Escape attract/demo texture validation pass, since this failure showed up as missing/garbled textures rather than an obvious boot failure.

PR #289, opened 2026-09-01
EMI to overlay extract, cached/mapped lookup table, "resident-occupant" memo

kerokline:
TLDR: For Breath of Fire 3 and 4, the .EMI layout is the game's own decomposition; runtime loses the relational grouping and leaves you reverse-engineering it later.

---

The framework's usual overlay bringup records overlays while someone plays. For most titles that's the only option. For Breath of Fire, and potentially others, there is a another route in the .EMI files.

Capture records overlays as they're hit: an executed address, the bytes resident at the time, reached along whatever path the player walked. Thread that across a playthrough of an RPG — town, field, encounter, menu, shop, dungeon, boss, and every branch between — and you accumulate a long, tangled chain of entrance points strung along the game's flow. It's a real record of what happened. But every node in it is keyed by address and play-path, never by module. You'd know the game went from one segment to the next; you would not know that these forty entry points are all one area script, those nineteen are one character's module, these belong to the shop system. The reason segments belong together is the one thing capture can't see — it's metadata in the disc's archive layout, gone the instant the bytes land in RAM.

Three things make that especially costly here: the branching factor is enormous, address reuse is aggressive (181 different area scripts share a single RAM address, indistinguishable to capture), and coverage is unprovable — you can never be sure you visited every optional area. The result is behavior without architecture: a trace that the game did these transitions, but no answer to "what are its subsystems?" — the exact unit you need for modding, performance work, or extension.

Breath of Fire III's .EMI containers already carry that answer. Each one's table of contents states every section's RAM destination, and the sections are stored contiguously — so the overlay set, and its grouping, is readable offline as a build input.

Compiling from that flat list initially introduced a good bit of latency while transitioning between gamestates, hence the need for the hashmap. The initial idea, steps taken, and per-session rationale, are covered in the attached  OVERLAY_EXTRACTION.md 

[OVERLAY_EXTRACTION.md](https://github.com/user-attachments/files/31672868/OVERLAY_EXTRACTION.md)
§§8–12.

Blast radius: static-overlay path only. The runtime change (aa6fa2c9) is confined to psx_overlay_static_code_matches(), which is unreachable without PSX_HAS_OVERLAY_DISPATCH; both codegen changes (69d783f5, 70153175) are confined to generate_overlay_dispatch(), emitted only under --static. The DLL-loader capture path used by default bringup method is not modified — capture-only titles compile and dispatch exactly as before.


kerokline:
Left in draft so you could review - Its an alternate overlay derivation path, but it shouldn't impact the older path, its just if someone finds it would work for whatever game they're working through. In draft for the moment while I get further in the game (to see if performance stays steady)

mstan:
Thanks for the PR! So I gave this one an eye, and going to try and summarize the intent as best I can in human terms. So correct me if I'm wrong on it.

Overlays make static recompilation a bit of a pain in the ass since they're basically different per game. Fortunately, Breath of Fire games are a bit more identifiable upfront, and can actually have some level of static analysis done, so this is an opt-in for these games (and games that share the methodology) where-in they're actually being statically compiled, avoiding the need to natively shard them? For any missed shards, developers that don't opt in, or overlays that get missed _despite_ opt in, the continued native sharding path will work as it always has.

Is that a fair representation of what this?

I did a bit of regression testing on a few titles that weren't able to take advantage of this, and it doesn't seem to compromise their native sharding, so I don't see this as a risk by any means. 

In a perfect world, we could eventually identify the various overlay methods and move them all towards static analysis and eventually find native sharding to be inert. This seems like a good step towards that -- assuming I've assessed this correctly. 

Do let me know if I've missed the mark here. Thanks!

kerokline:
> Is that a fair representation of what this?

Yup! I had already spent some time working through the format to figure out the pixel sheets, so I already had the header/section info figured out, and there have been other people that have taken time decoding the EMI files before, so I'm sure I gave a big nudge to Claude to try statically compiling what it could. 

Even for this game, not everything can be statically compiled, so its kind of a "both/and" approach. 

The real benefit is that it allows pre-knowledge of where classes of overlays are coming from, so game extensions and mods gain a lot of power / extensibility.

I'll move it to ready to review - thanks for the regression check!

PR #313, opened 2026-09-03
gpu: polyline terminator never in the first two vertices

kerokline:
The GP0 polyline parser tested every streamed word against the 0x55555555 terminator mask ((w & 0xF000F000) == 0x50005000). Hardware, per both oracles (Beetle mednafen/psx/gpu.c INCMD_PLINE, DuckStation HandleRenderPolyLineCommand / DrawingPolyLine), consumes the first two vertices unconditionally and afterwards tests only the first word of each vertex unit: the vertex word for mono polylines, the colour word for shaded ones. Shaded vertex words are never tested.

Psy-Q leaves the top byte of LINE_G* colour words as junk, so a legitimate colour such as 0x52545454 matched the mask at the second vertex and ended the polyline early. The remaining words were re-parsed as new commands and a later colour word, 0x02010101, executed as a 341x341 FILL. In Breath of Fire III (SLPS-00990) the item-use effect's shaded polylines hit this every time: the fill covers the terrain texture page at VRAM (448,256) and the whole ground vanishes for the rest of the area (terrain renders as CLUT entry 0, i.e. transparent).

The vertex count is folded into the existing polyline_has_prev int (mono: 0..2 vertices seen; shaded: 0=need V0, 1=need C1, 2=need Vn, 3=need Cn with n>=2) so the GPU savestate section keeps its size.

Verified on Breath of Fire III: loading the reproducing savestate on the old build zeroed the page 35 frames in; on this build it stays intact over 6000+ frames and the terrain renders.

### DIFF
--- a/runtime/src/gpu.c
+++ b/runtime/src/gpu.c
@@ -5360,19 +5360,31 @@ static void gpu_write_gp0_body(uint32_t val) {
         return;
     }
 
-    /* State: mono polyline — each word is a vertex (or terminator) */
+    /* Polyline terminator rule (Beetle mednafen/psx/gpu.c INCMD_PLINE,
+     * DuckStation gpu.cpp HandleRenderPolyLineCommand/DrawingPolyLine):
+     *
+     *  1. A polyline always has at least two vertices. The words of the first
+     *     two vertices are consumed unconditionally — mono [V0][V1], shaded
+     *     [V0][C1][V1] — and are NEVER tested for the terminator.
+     *  2. From the third vertex on, only the FIRST word of each vertex unit
+     *     is tested: the vertex word itself for mono, the colour word for
+     *     shaded. Shaded vertex words are never tested.
+     *  3. The test is (word & 0xF000F000) == 0x50005000 (0x55555555).
+     *
+     * Testing every word is wrong in a way games actually hit: Psy-Q leaves
+     * the top byte of LINE_G* colour words as junk, so a colour such as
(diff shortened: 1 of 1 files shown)


kerokline:
That above description is a little hard to parse, here's the summary ->

Same savestate (file slot04): old build-dbg after the wipe vs. fixed build-relprof several thousand frames later.

image
image

Root-caused and fixed. It was a framework GPU bug, not a texture-upload or savestate problem.

**What the glitch is.** The ground tiles are still drawn every frame, but their 8-bit texture page at VRAM (448,256) had been zeroed. Texel 0 maps to CLUT entry 0, which is transparent, so the whole ground renders as nothing. That is why it survived fleeing the battle: the field uses the same page until you leave the area.

**Why it happens.** The item-use effect draws gouraud-shaded polylines. Our GP0 parser tested every streamed word against the polyline terminator mask, and one of the effect's colour words, 0x52545454 with Psy-Q junk in its top byte, matched. The polyline ended early, the leftover words were re-parsed as new commands, and a later colour word 0x02010101 decoded as a 341x341 VRAM fill that landed right on top of the terrain texture. Both oracles, Beetle and DuckStation, never test the first two vertices and afterwards test only the first word of each vertex unit. The full evidence chain is in gpu-polyline-terminator.md.

**Fix and verification.** One change in the framework's gpu.c. I rebuilt build-relprof and loaded the glitched state headless: the texture stayed intact for over 6000 frames and the battle went on normally, versus a wipe 35 frames after load on the old build. The two screenshots above are the before and after.

PR #327, opened 2026-09-06
Preserve interlaced fields when drawing GPU primitives

Alexbeav:
When 480i drawing prohibits the active display area, raster primitives currently overwrite both row parities. A guest field-detection loop can then stay black at startup. Preserve the selected parity through native-row clipping, restore the caller clip and re-arm precision/perspective metadata on each submission. Fills, copies and transfers remain unmasked. The facade counts each original perspective triangle once, so row metadata re-arms do not inflate diagnostics. No title detection or guest code changes.

Fork review: https://github.com/Alexbeav/psxrecomp/pull/22. Cubic completed exact-head check 101479562581 with zero new issues. All earlier review jobs are terminal and findings are resolved in the final source or explicitly reconciled in the replies. The public branch is identical to this reviewed head.

Base `155e269ba5d50fe36523846f908cf124dc1561a0`; tested/reviewed head `6fcfb8adbdcd666a2fc8ba419dd4e2aa7ee1faaa`; five focused commits; 11 changed files:

- `docs/GPU_INTERLACED_RASTER.md`
- `runtime/CMakeLists.txt`
- `runtime/include/gpu.h`
- `runtime/include/gpu_interlace.h`
- `runtime/include/gpu_render.h`
- `runtime/include/gpu_sw_renderer.h`
- `runtime/src/gpu.c`
- `runtime/src/gpu_render.c`
- `runtime/src/gpu_sw_renderer.c`
- `runtime/src/gpu_vk_renderer.c`
- `runtime/tests/test_gpu_interlace_render.c`

Runtime files implement the field rule and facade-owned counter; backend metadata setters remain side-effect-free with respect to counting, and backend wide-overlay policy remains unchanged outside field masking; the C fixture tests mode/parity, all nine primitive families, native/4x output, clip restoration, transfers and metadata consumption and logical-triangle counts; CMake registers it; the document describes the invariant and prior behavioral evidence. Source-owned synthetic data only.

Validation on this review head, Windows x64 / WinLibs GCC16.1:

- Runtime CTest: 69 enabled pass; 2 pre-existing disabled. The initial review head fails 8 added counter assertions. The pre-bounds head fails exactly 2 bounded-row checks while pixel comparisons pass. Both fields at native/4x match unmasked pixel oracles even when precise vertices cross integer bounds. A tiny precise triangle submits 3 row clips instead of 256. A backend that intentionally ignores the clip reproduces 40 wrong field rows on the previous facade. Field-masked flat rectangles now submit one-row geometry, while the ordinary overlay geometry and backend policies stay unchanged. Both parities, no duplicate writes and clip restoration pass. This is a backend-contract fixture, not Vulkan hardware pixel qualification. Focused command: `ctest --test-dir <runtime-build> -R gpu_interlace_render_test --output-on-failure`.
- Recompiler suite:58/61 enabled pass;3 disabled. The three failures also reproduce from pristine upstream source: `dirty_text_continuation_guards` expects an absent source fragment; `release_zip` expects LF bytes on Windows; `aot_overlay_discovery` fails its hardcoded MSYS gcc subprocess. Recompiler sources are unchanged by this PR. Python tests ran with PYTHONUTF8=1.
- Owned Pro Pinball: Timeshock! USA SLUS-00639: fresh process through game frontend, start player game, launch ball, flipper/ball interaction and normal TCP quit at frame 6925. Exact build identifies this head. Digital pad, software/native, existing owned generated game/SCPH5552 code reused unchanged; no emitter-regeneration claim. The title is a private local port, so no public game source or copyrighted payload is attached.
- Repeat evidence: unchanged exact executable also completed a second fresh player route at frame 6923. One earlier attempt on this head stayed in the startup pixel-test loop through frame 6925 and then exited by TCP request. Its cause is unestablished; a prior-head comparison passed, and the exact edge-pixel synthetic case passes. Do not read the successful repeats as proof that startup is free of intermittent failures. The private staging failure in a separate control never launched a runtime and is excluded.
- Current upstream runtime does not compile with public recomp-ui/master773155ae: unrelated new lobby/spectator fields are missing. Retail qualification therefore uses supported `PSX_RECOMP_UI=OFF`, direct runtime launch into the game frontend. This is not launcher compatibility acceptance.

Field selection uses the existing vertical-blank latch, not scanline-accurate timing. Row clipping can increase hardware submissions; the separate https://github.com/mstan/psxrecomp/pull/326 addresses GL readback volume. No Vulkan-wide, netplay, full-game, all-audio or save/load acceptance is claimed. Runtime-only consumer pin update is required; no game regeneration is required by this correction.

### DIFF
--- a/docs/GPU_INTERLACED_RASTER.md
+++ b/docs/GPU_INTERLACED_RASTER.md
@@ -1,0 +1,42 @@
+# Interlaced primitive drawing
+
+The GPU preserves one row parity while drawing in 480-line interlaced mode
+when GP0(E1h) prohibits drawing to the display area. The display origin and
+current field select that parity. Progressive modes and the draw-to-display
+override draw both parities. Fills, copies and VRAM transfers keep both fields.
+
+`gpu_raster_skipped_row()` reads current GPU state for each primitive. The
+renderer facade applies the field mask with native-row clip rectangles and
+restores the original clip. This also preserves the corresponding rows in
+scaled surfaces. Each clipped triangle retains its precision and perspective
+metadata, which hardware backends otherwise consume on the first submission.
+No title address or title detection selects this behavior.
+
+Pro Pinball: Timeshock! USA SLUS-00639 repeatedly draws and reads a test pixel
+at (1020,511) during startup. Frozen source 63446a28 always returned the drawn
+pixel and stayed black. This correction lets the guest's existing field test
+finish. The same disc, BIOS and settings then reach the table.
+
+The field latch uses the runtime's existing vertical-blank model. This change
+does not claim scanline-accurate timing. Row clipping can increase submissions
+for large interlaced primitives; no general performance claim is made.
+
+## Verification
+
+Configure `runtime/` with `BUILD_TESTING=ON`, then run
+`ctest --test-dir <build> -R gpu_interlace_render_test --output-on-failure`.
+The fixture covers the mode predicate, origin/field parity, all nine primitive
+families, clip restoration, transfers, mode changes, scaled output and
+triangle metadata consumed by a backend. The original renderer fails the
+same row-preservation assertions.
+
+## Prior art
+
+- [PSX-SPX GPU specification](https://psx-spx.consoledev.net/graphicsprocessingunitgpu/)
+  describes interlace affecting draw commands.
+- [DuckStation difficult games](https://github.com/stenzek/duckstation/wiki/Difficult-to-Emulate-Games)
+  identifies required line skipping for Pro Pinball startup.
+
+These sources supplied behavioral evidence. The implementation and regression
+fixture were written independently. Exact title run receipts and route limits
+are maintained by the Timeshock correction project.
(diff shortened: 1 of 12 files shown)


mstan:
@Alexbeav I've opened #336 to roll back this PR in commit 24c6c56c, after tying the retail SCPH1001 BIOS-screen stall to `0661ce77` with matched Ape Escape builds.

- Exact parent `155e269b`: retail LLE boot reaches game code; normal close at frame 2192.
- `0661ce77` alone: severe Sony-screen stall around frames 375-376.
- Master immediately before #334 boots; master after #334 reproduces it.
- All comparisons used `PSX_BIOS_HLE=0`, the same generated game/BIOS code, BIOS/disc, config, and OpenGL settings. This is real retail intro boot, not HLE shell skip.

The per-row primitive resubmission/clip changes amplify an existing malformed DMA list into enormous GL work. A separate CPU/DMA ownership experiment fixes the packet race, but it needs VHD/Spot compatibility coverage, so it is not bundled into this rollback.

#336 reverts only #327's five commits and preserves #326/#328. The rollback reaches Ape's title/menu and frame 11669 on retail BIOS; OpenBIOS also boots. MMX6 retail boot passes and closes normally at frame 4123. Eight focused tests pass, and the retained GL readback fixture passes all 71 checks at 1x and 4x.

This temporarily restores the previous interlaced-field limitation. A replacement should avoid per-scanline backend draw/flush multiplication and include windowed retail SCPH1001 boot in validation. Full attribution and scope are recorded in #336's investigation document.

PR #353, opened 2026-09-13
feat(OpenGL renderer): add post presentation gamma control

### LINKED ISSUE #352: Expose a presentation gamma/brightness/contrast API for mods in `gpu_gl_renderer.c`
Hello,

I'm working on a **Wild 9** recompilation using `psxrecomp`, and I think being able to adjust the final presentation brightness from a mod would be very useful on this game, and perhaps others.

In my local copy of `gpu_gl_renderer.c`, I currently have a small set of changes that exposes a function such as:

```c
void gl_renderer_set_post_gamma(float gamma);
```

This applies a gamma curve in the final OpenGL presentation stage, after the game frame has been rendered. In practice, this allows me to lift the very dark areas of the game without modifying its textures or generated/recompiled game code.

The current OpenGL renderer already has a final presentation shader (`PRESENT_FS`) and a presentation path through `present_target_quad()`, so this seems like a natural place for this kind of post-processing control.

For example, a mod could potentially do something like:

```c
gl_renderer_set_post_gamma(1.30f);
```

and expose different brightness levels as mod options.

I think this could also be useful beyond my particular recompilation, since other games may benefit from simple display adjustments such as gamma, brightness, or contrast without requiring game-specific patches.

Would you be interested in exposing an API for this in `psxrecomp`, or do you already have plans to add something similar?

If this is something you would consider upstreaming, I'd be happy to make an initial implementation and open a PR, assuming the general direction is acceptable first.

Thank you.

plinkr:
Implements the presentation gamma control proposed in #352.

This adds a public `gl_renderer_set_post_gamma(float gamma)` API to the OpenGL renderer, allowing mods and external plugins to adjust the final display gamma at runtime without modifying guest rendering state, textures, or generated/recompiled game code.

The adjustment is applied in the final OpenGL presentation stage, after the game frame has been rendered.

### Behavior

* `gamma = 1.0` preserves the default presentation.
* Values greater than `1.0` lift shadows and mid-tones.
* The adjustment applies only to game content and temporal interpolation passes.
* Host UI elements, bezels, and previously composed frames are left untouched.
* Changing the gamma value invalidates the presentation buffer so the new value becomes visible immediately, without requiring additional guest video memory writes.

### Plugin usage

Plugins can now expose game-specific brightness or visibility adjustments without modifying generated game code:

```c
#include "mod_plugins.h"

/*
 * Example - Final-frame brightness
 *
 * The actual effect lives in the OpenGL present shader. The plugin only
 * selects the runtime post-process value, so no generated game code is
 * modified.
 *
 * gamma = 1.0 -> vanilla
 * gamma > 1.0 -> lifts shadows/mid-tones
 */
#define POST_GAMMA 1.30f

extern void gl_renderer_set_post_gamma(float gamma);

static void brightness_activate(void)
{
    gl_renderer_set_post_gamma(POST_GAMMA);
}

PSX_MOD_CONSTRUCTOR(register_brightness_plugin)
{
    (void)psx_mod_register_activation_plugin(
        "brightness",
        brightness_activate);
}
```

This is useful for titles or enhancement mods where the original presentation is excessively dark, while keeping the adjustment entirely in the renderer's post-presentation stage.

Closes #352.

### DIFF
--- a/runtime/include/gpu_gl_renderer.h
+++ b/runtime/include/gpu_gl_renderer.h
@@ -130,6 +130,15 @@ void gl_renderer_set_display_aspect(int num, int den);
 void gl_renderer_set_scanlines(int on, float strength);
 int  gl_renderer_get_scanlines(float *strength);
 
+/* Presentation-only gamma adjustment. gamma = 1.0 is the identity; values
+ * above 1.0 lift shadow detail and values below 1.0 darken it. The adjustment
+ * is applied once to game content in the final GL presentation pass, including
+ * temporal interpolation, but not to the bezel, host OSD, black margins, or an
+ * already-composed hold-last image. Non-finite and out-of-range values are
+ * clamped to a safe range. Safe to call before GL context creation. */
+void  gl_renderer_set_post_gamma(float gamma);
+float gl_renderer_get_post_gamma(void);
+
 /* Select full native-wide mirror rendering instead of the centre-splice fast
  * path. Textured edge expansion needs the complete mirror surface. */
 void gl_renderer_set_wide_fast(int on);
--- a/runtime/src/gpu_gl_renderer.c
+++ b/runtime/src/gpu_gl_renderer.c
@@ -320,6 +320,11 @@ static void          gl_swap_with_osd(void);
 static GLint         s_present_uTex = -1, s_present_uUvRect = -1;
 static GLint         s_present_uTexSize = -1, s_present_uSharpScale = -1;
 static GLint         s_present_uSharp = -1;
+static GLint         s_present_uGamma = -1;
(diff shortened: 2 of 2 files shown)


mstan (approved):
No blocking findings. The default gamma path is an exact identity branch, every
shared presentation-program draw site explicitly selects game gamma or identity,
and the setter safely normalizes/clamps values and invalidates cached presentation
state without requiring a live GL context.

Validation passed: C11 syntax check; a ThinLTO-linked source-plugin probe for
default, NaN, clamped and nonidentity values; and combined exact-BIOS title runs
in Tomba, Mega Man X6 and Ape Escape. This is low regression risk.

PR #356, opened 2026-09-13
experiment(build): opt-in runtime IPO and PGO validation gates

mstan:
## Experimental opt-in runtime IPO

Adds `PSX_RUNTIME_IPO`, default OFF. An explicit request checks C/C++ compiler
support and enables IPO only on optimized runtime target configurations, not
all project/third-party targets or the standalone overlay compiler.

The PE follow-up reports near-60 FPS with project-wide LTO plus retrained PGO.
That result is not independently reproduced here and is not identical in scope
to this narrower option. No runtime timing/check semantics or shipping defaults
are changed. This does not implement the IRQ-batching prototype.

Validation:

- Real Clang 22 C/C++ IPO compile/link/run, OFF/default and inherited-policy,
  unsupported-toolchain, unrelated-target isolation and Debug-OFF checks pass.
- Existing mocked PGO generate/train/use success and failure paths, plus the
  unprofiled overlay DLL command check, pass.
- Clang 22.1.8 ThinLTO builds ran Tomba, Mega Man X6 and Ape Escape for more
  than 11,000 frames each with LLE boot, native overlay dispatch, nonzero SPU
  output, zero audio underruns and no kernel-state mismatch. Representative
  screenshots showed normal title output.

The title runs establish a correctness baseline for this opt-in experiment.
They are not real PGO training or performance measurements and do not justify
enabling IPO by default.

See `docs/internal/RUNTIME_IPO_EXPERIMENT.md` for exact scope, remaining
performance/platform evidence, and promotion gates.

Tracking: `beads-eio.3.149`.

### DIFF
--- a/cmake/psx_runtime_ipo.cmake
+++ b/cmake/psx_runtime_ipo.cmake
@@ -1,0 +1,29 @@
+include_guard(GLOBAL)
+
+option(PSX_RUNTIME_IPO
+    "Experimental IPO/LTO for optimized runtime targets (requires supported C/C++ toolchain)"
+    OFF)
+
+function(psxrecomp_apply_runtime_ipo target)
+    # OFF must not alter a parent project's explicit IPO policy or add flags.
+    if(NOT PSX_RUNTIME_IPO)
+        return()
+    endif()
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT _psx_ipo_ok OUTPUT _psx_ipo_error LANGUAGES C CXX)
+    if(NOT _psx_ipo_ok)
+        message(FATAL_ERROR
+            "PSX_RUNTIME_IPO=ON was requested, but C/C++ IPO is unsupported.\n"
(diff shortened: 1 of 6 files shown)


mstan:
Independent adversarial review found no blocker. The option returns without
touching IPO state when OFF, scopes IPO to the runtime target and optimized
configurations when ON, and fails explicitly on an unsupported requested
toolchain.

Validation passed with Clang 22.1.8: default/inherited policy, unsupported
failure, scoped real IPO compile/link/run, Debug-OFF isolation, mocked PGO flow,
and exact-BIOS ThinLTO runs beyond 11,000 frames in Tomba, Mega Man X6 and Ape
Escape. All title runs had zero kernel-state mismatches and audio underruns with
normal output.

Recommendation: merge as a low-risk experimental option that remains OFF by
default. These results do not establish a performance win or validate real PGO
profile collection.

PR #357, opened 2026-09-13
RFC: FMV IRQ batching correctness gates and counterexample

mstan:
## FMV IRQ batching RFC and correctness guard

Preserves all production behavior and records the review gates for the supplied
WO-8–WO-11 FMV briefs. The actual contributor prototype/diagnostic patches and
PE capture bytes are not available, so this deliberately does not reconstruct
an unverified implementation from prose.

Adds a native counterexample against the real I-cache source: retaining header
fetches while omitting a cold interior fetch changes cycles (14 -> 7), load
give-back (0 -> 99) and the body cache tag. The native test passes. It tests the
described elision operation; it does not claim to execute the missing patch.

`docs/internal/FMV_IRQ_BATCHING_RFC.md` records guest-state/deadline/CPS/cache
invariants, synthetic and title/oracle validation requirements, and the exact
supplemental evidence needed. It also explains why decoder profiling, loader
diagnostic history and PGO attribution remain separate questions.

The confirmed structural CFG repair is #354; optional build repair is #355.
There are no experimental IRQ flags, production source modifications or
performance claims in this RFC. It is ready to merge as documentation plus a
regression guard, not as an IRQ batching implementation.

Tracking: `beads-eio.3.150`.

### DIFF
--- a/docs/internal/FMV_IRQ_BATCHING_RFC.md
+++ b/docs/internal/FMV_IRQ_BATCHING_RFC.md
@@ -1,0 +1,97 @@
+# Draft RFC: FMV callback overhead without guest-state changes
+
+Status: design review only. No IRQ batching implementation or runtime flag is
+added. Tracking: `beads-eio.3.150`. The contributor's actual prototype patch,
+instrumentation patch, PE capture bytes and raw measurements are not available.
+
+## What is established
+
+WO-10 correctly identifies repeated generated block callbacks. With precise
+slicing disabled, the runtime slice body returns early, but the DLL still calls
+through its callback shim. That observation does not quantify the host cost.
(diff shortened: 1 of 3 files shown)


mstan:
Independent adversarial review found no blocker. The native guard uses the real
I-cache implementation and validly demonstrates that omitting the cold body
fetch changes guest cycles (14 to 7), load give-back (0 to 99), and cache-tag
state. The test passes and no production path changes.

Recommendation: merge as a low-risk RFC and correctness guard. It is not an IRQ
batching implementation or evidence of a performance fix; the RFC accurately
keeps the wider cache, deadline, CPS and title-oracle cases as requirements for
any future implementation.

PR #369, opened 2026-09-16
setup: say why disc verification failed instead of "wrong dump"

Alexbeav:
## Symptom

A player picks a `.chd` in the launcher's disc picker, the runtime mounts it fine, and first-run setup then stops with:

```
Disc verification failed (wrong dump).
```

The dump is correct. Reported against three Wave 3 kits by players who then went looking for a bad rip.

## Cause

`psxrecomp_cli.py` `verify_disc_path` hashes the disc against the uncompressed track digests recorded in `[prepare_disc]`. It resolves a `.cue` to its `.bin`, but has no `.chd` path, so it hashes the compressed container. That can never match, and the mismatch is raised as a generic digest error.

`cli_fail_msg` in the codegen host then threw away whatever the CLI said for exit 3 and printed the flat "wrong dump" string, even though `cli_tail_note` had already captured the CLI's own error line.

## Change

- `verify_disc_path`: detect `.chd` and raise a `DiscVerifyError` that says the build tools verify against uncompressed digests, that the dump is not the problem, and gives the `chdman extractcd` command to produce a `.cue`/`.bin`.
- `cli_fail_msg`: for exit 3, keep the CLI's captured reason when there is one, and fall back to the old string only when there is not.

CHD generation stays unsupported. This only stops it being reported as a bad disc.

## Verification

- Original failure reproduced on a Wave 3 kit with `Metal Gear Solid (Europe) (Disc 1).chd`: the runtime mounted the disc and the host reported "wrong dump".
- `python -m py_compile psxrecomp_cli.py` on this branch.
- The host change is a two-line branch on the existing `CliTail` capture; not rebuilt end to end here.


Alexbeav:
Superseded by #370, which reads .chd through libchdr in the Python tools instead of only improving the message. Closing this one.

PR #370, opened 2026-09-16
setup: read .chd through libchdr in the Python disc tools

Alexbeav:
## Problem

The runtime mounts a `.chd` natively and the launcher lists it in the disc picker, so a player can select one and see a valid disc identity. First-run setup then fails: `verify-disc`, `tools/prepare_disc.py` and `probe_disc.py` hash and parse raw 2352-byte track data, which a CHD stores compressed. Nothing on the Python side could read it. Three Wave 3 reports traced back to this, and "install MAME for chdman" is not an answer for players.

## Change

libchdr is already vendored under `third_party/` and compiled for the runtime and the native media CLI. This builds it once more as a shared library in `recompiler/` (target `chdr`, output next to the emitters, static CRT on Windows so a plain `python.exe` loads it with no toolchain on PATH) and reads the CHD from Python through ctypes.

- **`tools/psx_chd.py`** (new): ctypes binding plus track-table logic mirrored from `runtime/src/iso_reader.cpp`: 2448-byte frames, 4-frame track padding, stored versus virtual pregaps, big-endian audio. It reproduces the `.bin` files a Redump dump of the same disc contains, in either of Redump's layouts (one bin per track, or one bin for the whole disc), and hashes both in one pass so the caller can match `[prepare_disc]`.
- **`psxrecomp_cli.py`**: `verify-disc` and `generate` hash the reconstructed track bytes, never the container. `ensure_chd_reader` builds target `chdr` on demand for a kit whose emitters predate it and falls back to an extract-it-first message that says the dump is not the problem.
- **`tools/prepare_disc.py`**: a `.chd` source is written into `out_dir` in whichever layout the recorded digests match, then the normal cue path runs in place, so the data track is written once, SBI companions are looked up next to the CHD, and the receipt names the CHD the player chose.
- **`tools/new_project_layout/probe_disc.py`**: accepts a `.chd`, extracts to a scratch directory, probes the cue.
- **`recompiler/CMakeLists.txt`**: scoped `BUILD_SHARED_LIBS` around the libchdr include; codecs stay static, `chdr-static` is untouched.
- **`docs/config_schema.md`**: the "not .chd" sentence is now wrong and is replaced.

Cooked track types (MODE1, MODE2_FORM1, ...) are refused with a clear message rather than guessed at; PS1 dumps are MODE2_RAW plus AUDIO.

## Verification

Real discs, from a Redump library converted with chdman:

- `007 - The World Is Not Enough (USA).chd`, one MODE2_RAW track: reproduces the catalogued size, MD5 and SHA-1 exactly.
- `Alien Resurrection (Europe).chd`, MODE2_RAW plus three AUDIO tracks with stored 150-frame pregaps: track 1 reproduces its catalogued digests exactly. `psxrecomp_cli.py verify-disc` accepts it and `tools/prepare_disc.py` stages all four tracks, extracts `SLES_029.13`, and writes the receipt, exit 0.

Tests (`tools/tests/test_psx_chd.py`, 23 pass) build a synthetic uncompressed v5 CHD around a cooked ISO 9660 image, so no retail data is involved. They cover metadata parsing, layout, cue rendering, byte-exact read-back with audio swap, both extraction layouts, `verify_disc_path` matching either layout, the no-reader message, `prepare_disc` for multi-track and single-track sources, the refusal on unknown digests, and `probe_disc`. The eight decode tests skip with a reason when no libchdr is built.

Built and exercised with MinGW (WinLibs) on Windows. The `chdr` DLL depends only on kernel32 and the UCRT.

## Notes

- Overlaps #369 on the `verify_disc_path` lines; if that merges first the conflict is the removed `.chd` message branch. #369's host-side change (keep the CLI's reason on exit 3) remains useful on its own.
- `psx_chd.py` is Python 3.9 compatible, as the other tools are.


Alexbeav:
macOS result, Intel MacBook Pro (Ventura 13.7.8, Python 3.9.6, Apple clang 15, cmake 4.4.3 + ninja):

- `cmake -S recompiler -B recompiler/build -G Ninja` configures and `--target chdr` builds `libchdr.0.3.dylib` (with the `libchdr.dylib` symlink) into the build dir; `otool -L` shows only `libSystem.B.dylib`.
- `python3 -m unittest tools.tests.test_psx_chd`: 12 tests, OK, on Python 3.9.6.
- `Alien Resurrection (Europe).chd` from the same Redump library: track 1 reproduces size 503850144, MD5 `e4aa0eb1…`, SHA-1 `5642c4f4…` exactly; 35 s over SMB on two cores.

`8cb24ee1` fixes one test that compared temp paths across the `/var` to `/private/var` symlink. Linux laptop check pending.

Alexbeav:
Linux result, Dell G15 laptop (Rocky Linux 9.8, kernel 5.14.0-687.46.1, Python 3.9.25, GCC, cmake 3.31.8 + ninja):

- `cmake -S recompiler -B recompiler/build -G Ninja` configures and `--target chdr` builds `libchdr.so.0.3` with the `libchdr.so` and `.so.0` symlinks; `ldd` shows only libc.
- `python3 -m unittest tools.tests.test_psx_chd`: 12 tests, OK, on Python 3.9.25.
- `Alien Resurrection (Europe).chd` over a CIFS mount: track 1 reproduces size 503850144, MD5 `e4aa0eb1…`, SHA-1 `5642c4f4…` exactly, 18.7 s.

That completes Windows (MinGW), macOS (Intel, Apple clang) and Linux (GCC) for the shared library, the tests, and byte-exact reconstruction of the same real disc on all three.

PR #373, opened 2026-09-16
runtime: spawn overlay autocompile on POSIX hosts

Alexbeav:
## Problem

`autocompile_request()` returned 0 on every non-Windows host (`/* non-Windows hosts: manual compile flow only */`). Captures were written, a staged toolchain was found, the boot log said "overlay autocompile enabled", and not one shard was ever compiled at runtime: a Linux or macOS player ran every overlay interpreted for the whole session. Measured on Parasite Eve, Rocky 9, gcc beside the exe: 1.35e9 interpreted instructions by frame 13k, 0 shards.

A second, smaller issue in the same file: a 208-address `--force-interior` list made the runtime-built command 6,392 chars, `snprintf` truncated it mid-argument into a 4096-byte buffer, every compile failed identically and autocompile latched degraded for the process lifetime with nothing pointing at the cause.

## Change

- The command buffer is 8192 bytes (cmd.exe's own ceiling), the wrapper is sized from it, and an over-long command is refused loudly at configure time instead of truncated.
- A POSIX spawner mirroring the Windows path: `/bin/sh -c <cmd>` in its own process group at nice 10, stdout and stderr through one pipe into the same tail ring with the streaming `PSX_SHARD_RESULT` parse, a pthread watcher that drains the pipe and reaps the child, and the emulation thread's poll turning DONE into the one idempotent cache rescan. No `PSX_SHARD_PUBLISHED` handoff off Windows: the rescan is the fallback the Windows path already takes when a marker line is cut off.
- Shutdown SIGTERMs the group and joins; `PR_SET_PDEATHSIG` on Linux plus the dying pipe cover a runtime that never reaches shutdown.
- The `PSX_OVERLAY_*` environment pins are set once at configure time (glibc `setenv` is not safe against a concurrent `getenv`). Configure names the interpreter it will run and marks the run degraded when that path is not executable.
- `shard_result_values`, the child-line buffer and `publish_parse_locked` move out of the Windows block; `child_line_locked` is per-platform. `Threads::Threads` is linked unconditionally on Unix.

## Verification

Rocky 9, Parasite Eve diagnostic product, empty cache, 150 s headless boot:

| probe | result |
|---|---|
| shards compiled | 0 → 68 |
| interpreted instructions by frame ~14.5k | 1.35e9 → 0.39e9, flat from frame ~4k |
| debug server `autocompile_status` | run 1 `shard_ok=5 rescans=1 last_exit=0`; run 2 observed `running` with python, gcc and cc1 in the tree |
| runtime exit mid-compile, and SIGKILL mid-compile | no orphans, no zombies |
| Windows | `autocompile_publication_test` and `autocompile_degraded_test` pass after the refactor; `-fsyntax-only` clean with the release flags |


Alexbeav:
macOS verification (Intel Mac, macOS 13, 8 GB; Parasite Eve diagnostic product, empty cache, headless):

| probe | result |
|---|---|
| shards compiled | 9 in the first 24 s (`cache/.../gcc/macos-x64/...`), counters over the debug server: run 1 finished, run 2 running |
| runtime exit mid-compile | exit 0, no compile processes left |
| SIGKILL of the runtime mid-compile | no `PR_SET_PDEATHSIG` on macOS; the compile tree exits on the broken pipe within 10 s, none left afterwards |

Getting a toolchain beside the exe on macOS needed a separate fix, now on #376: `release_stage.py` had no macOS toolchain pin.

PR #376, opened 2026-09-16
cli: stage the overlay toolchain beside every built product

Alexbeav:
## Problem

`overlay_cache = true` only enables the loader; native shards still need a toolchain beside the exe. `tools/release_stage.py` can stage one, but nothing in the player's path called it: no setup package ever shipped an `overlay_toolchain/`, and no CLI rebuild ever staged one, so the runtime's toolchain gate was false on every player build and every streamed overlay ran interpreted. Where a compiler was present, `autocompile_toolchain_available()` only probed PATH, so a Windows player with the wizard's clang pack but nothing on PATH fell to tcc shards or to nothing.

## Change

- `stage_overlay_toolchain_for_product()` in `psxrecomp_cli.py`, called after the rebuild resolves the product: pinned Python + `compile_overlays.py` + this pin's recompiler + headers + `overlay_codegen_hash.h`, through `release_stage.stage_toolchain`. Best effort: a staging failure leaves the product playable with overlays interpreted and says so.
- It writes `overlay_toolchain/compiler.txt` naming the compiler the CLI already has (the wizard's clang pack on Windows, the native cc on POSIX). `main.cpp` reads that file, counts it as an available gcc tier, and passes it as `--gcc` to the compile command.
- `release_stage.py`: the Windows DLL copy tolerates an emitter directory without MinGW runtime DLLs (a statically linked recompiler) and takes DLLs only from the emitter's own directory, never from PATH (Git's mingw64 on PATH carries the MSVCRT runtime, which was the 0xC0000139 trap).

## Verification

- Nightmare Creatures II kit, Windows, PATH stripped of every compiler: toolchain staged beside both products, `compiler.txt` names the pack's clang, the runtime logs the gcc tier "with <that clang>", 4 shards compiled in the first minute.
- Parasite Eve kit, Rocky 9: toolchain staged (69 MB, 3961 files), `compiler.txt` names `/usr/bin/gcc`; with #373 the runtime compiled 68 shards in 150 s.


Alexbeav:
Added a commit: `release_stage.py` had toolchain pins for `win` and `linux` only and chose `linux` for every non-Windows host, so no Mac ever got a bundled interpreter (and the stock macOS 13/14 `python3` is 3.9, which cannot run `compile_overlays.py`). The commit adds `macos-x64` and `macos-arm64` pins, the same python-build-standalone 3.13.1 release as Linux, with SHA256 matched against the publisher's sidecar files, and picks the key from the host architecture.

Verified on an Intel Mac: the toolchain stages beside the product (45 MB), and with #373 the runtime compiled 9 shards in the first 24 s.

PR #386, opened 2026-09-19
tools: enrich runtime overlay captures with static roots

### LINKED ISSUE #380: Captures never record function entries, which costs CPS titles their region compiles
Hit this porting Ace Combat 3 (SLPS-02020) at `193a60b8`, Linux, gcc 16.2.1.

`overlay_capture.c:438` writes the field as a literal:

```c
fprintf(f, "    \"function_entry_pcs\": [],\n");
```

Nothing backs it — no bitmap, no loop. And `seeds` just repeats `dispatch_entry_pcs`, same loop over the same bitmap a few lines down. So a capture only ever carries dispatch entries.

That means `compile_overlays.py` always falls back to `_callable_legacy_seed` (`:1079`), which guesses at function starts from shape: `addiu sp, sp, -N` not preceded by control flow, or `jr ra` eight bytes back. Fine if dispatch entries land on prologues. AC3 is continuation-passing, so they don't — they're resume points in the middle of functions.

Out of 7779 dispatch entries across 160 overlay images, 357 pass (4.6%). Every one of the 7422 failures is "not a prologue"; the opcodes there are plain `LW`, `SW`, `ADDIU`, `LUI`. **104 of the 160 images end up with no walk roots at all**, hit the `SKIP: no shared walk-root seeds` at `:6369`, skip the region compile entirely, and fragment into one island per entry.

The islands then eat the file budget. `RUNTIME_CACHE_FILE_CAP` is 4096 (`:5006`), 3957 of my 4096 files were islands with a median coverage of 362 bytes, and once it filled up the next reshard did nothing:

```
PSX_SHARD_RESULT ok=0 failed=0 skipped=0 capacity_fastpath=160
```

All 160 images skipped before byte decode, exit 0. Candidate cap wasn't the problem — that was 8054 of 32768. So the cache jams and no amount of further play helps, while the `ok=` line still looks like success.

Declaring the entries in `game.toml` fixes it. `_collect_toml_overlay_entries` (`:1487`) already accepts them, so I derived them from the captured bytes — direct `JAL` targets, plus prologue-shaped addresses found by scanning every word instead of only dispatch entries. 10584 entries, no image left rootless. Same captures, rebuilt:

| | before | after |
|---|---|---|
| shards | 4096 (at cap) | 1646 |
| islands | 3957 | 955 |
| covered bytes | 4.2M | 9.8M |
| `F` entries served | 9070 | 21008 |

15 images still fail, but loudly and for real reasons — `UNSUPPORTED_INSTRUCTION` around the kernel vectors, one reserved branch encoding at `0x800CC0E0`.

Worth mentioning: I'd been chasing a garbled FMV for two sessions, convinced a specific function was miscompiling. It compiles fine inside a region. It was only ever wrong as an isolated island. I don't have a clean reduction for that divergence, so I'm not filing it separately yet.

Smallest useful fix is probably just warning when an image yields zero walk roots, instead of one line in a long log. Documenting the `game.toml` route for CPS titles would cover it too. Recording the entries properly seems doable — the runtime knows a call target when it dispatches one — but that's your call.

AC3 is the only game I've tried, so this might not generalize. I'm not a C programmer, either, so I can't maintain a patch. This is a report.

Most digging was done with an AI Assistant, Claude Opus 5. Line references & numbers do check out, and my limited C says these bugs are real, but double check. This framing is partly its work.

mstan:
Fixes #380 in the supported tooling layer without claiming runtime dispatches are function boundaries.

Adds `tools/enrich_overlay_captures.py`, a post-capture pass that unions conservative `static_discovery_entry_pcs` derived from captured bytes: framed entries, direct `jal` targets, and members of dense in-image pointer tables. Every non-frame candidate must pass the existing bounded CFG proof; `compile_overlays.py` validates the entries again. The runtime-owned `function_entry_pcs` field remains untouched because CPS continuations are also external dirty-RAM dispatches.

Validation:
- `tools/tests/test_enrich_overlay_captures.py` (3 tests)
- `tools/test_compile_overlays_static_split.py` (10 tests)
- CMake configure registration guard + `ctest -R enrich_overlay_captures`

Draft pending validation against a real rootless CPS capture, such as the reported AC3 data.


x6568tank:
Ran it against the ACE3 capture; it works, removes the cache-cap pressure, and adds no new failures, though my local fix finds more roots.

905 capture files merged -> 250 distinct images, 134 load addresses, 386795 executed PCs. Three runs, each a cold-compile into its own output directory, all three using the same `compile_overlays.py` and recompiler at `193a60b8` so enrichment is the only variable that could impact output. Set 1 is merged untouched, set 2 is `enrich_overlay_captures.py` from this PR, set 3 is my own `enrich_captures.py`, which derives the same field by walking `compile_overlays.py`'s own prologue logic.

| | no enrichment | #386 | mine |
|---|---|---|---|
| rootless images | 250 / 250 | 1 | 0 |
| static roots added | 0 | 10,191 | 15,008 |
| wall | 1307s | 841s | 1166s |
| `PSX_SHARD_RESULT` | ok=4096 failed=0 skipped=1928 | ok=1186 failed=8 skipped=84 | ok=2112 failed=19 skipped=219 |
| island shards | 159 | 76 | 41 |
| capacity-preflight skips | 3028 | 0 | 0 |
| entries served | 10,303 | 19,442 | 25,394 |
| covered bytes | 3,045,612 | 3,767,744 | 4,497,864 |
| cache on disk | 666 MB | 548 MB | 758 MB |

Rootless images shrink 250 > 1, islands more than halve, and enrich pass only costs 0.6s. 8 failures are all `fragment: no-func-ids`, and they all appear in my reshard's 19, so no new breaks.

Worth a note, unenriched run is the only one that exits 0, reporting `failed=0` while hitting capacity preflight 3028 times — and the two runs that build a good cache exit 2. Hard numbers backing up #380.

Yours does find 4817 fewer roots than mine, worth +5952 served entries, +731 KB covered, and -35 islands. Both sides use the same compile tooling, so that has to be the enrichment. Ship it anyway, I can't maintain mine if upstream code breaks it and a tool in `tools/` I don't have to rebuild every time prologue logic moves is worth more than those roots.

Couple things; `covered bytes` & `entries served` aren't reported figures, I summed them from the `.ranges` manifests keyed per-image CRC & de-duping ranges by start address, meaning I could be undercounting. Also only used the enrich tool from this PR, not its `compile_overlays` commit, since pinning the compile side was the point.

Happy to rerun if you wanna test anything else.

